PR #69, opened 2022-07-14
Add Google Tag Manager

alexfornuto:
Adds `docusaurus-gtm-plugin`. The GTM ID has been set in Netlify's environment variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@mui/x-data-grid": "^5.10.0",
     "@mui/x-data-grid-pro": "^5.10.0",
     "clsx": "^1.1.1",
+    "docusaurus-gtm-plugin": "^0.0.2",
     "dotenv": "^16.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "@mui/x-data-grid": "^5.10.0",
     "@mui/x-data-grid-pro": "^5.10.0",
     "clsx": "^1.1.1",
+    "docusaurus-gtm-plugin": "^0.0.2",
```
desimone, line 27: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-google-gtag is this different? Seems like the official one?

alexfornuto: It seems to me like there's a difference between "google site tag" and a "google tag manager". They each take different IDs assigned by Google. We already have gtag implemented and it didn't seem to do the job for @cmo-pomerium, maybe he can weigh in further.

desimone (approved):

PR #75, opened 2022-07-21
add bamboo and zenefits

wasaga:
Related issues: 
- https://github.com/pomerium/internal/issues/897
- https://github.com/pomerium/internal/issues/896

The docker images assume `Authorize` middleware removal, will be available once https://github.com/pomerium/datasource/pull/43 is merged.


content/docs/enterprise/external-data/zenefits.mdx
```diff
@@ -1,44 +1,112 @@
...
+You may add a similar policy around `type` field to i.e. differentiate between employees and contractors. 
 
 ## Reference
 
-[Docker Compose]: https://docs.docker.com/compose/
\ No newline at end of file
+The reference Zenefits data source provides the following record details:
+
+- Zenefits `workEmail` field becomes an `id` of each record, intended to be matched with Pomerium `user.email` foreign key.
+- Zenefits `status` and `type` are exposed as example, and you may build policies around the values of those two fields. See [Zenefits People API] for details.
```
alexfornuto, line 105: This makes it sound like `status` and `type` are only exposed in the example. What are you trying to say here, so I can massage the words?

alexfornuto (approved):
One item below that needs discussion.

content/docs/enterprise/external-data/zenefits.mdx
wasaga: I mean we're just scratching a surface of use cases that are possible for both Bamboo & Zenefits. They have much more information that potentially may be useful for Pomerium policies - i.e. custom fields, employee location, etc.

Issue #87, opened 2022-07-29
Dark mode tables are illegible


alexrudd2:
Page: https://docs.pomerium.io/docs/reference (or https://www.pomerium.com/docs/reference)

## What's incorrect or missing
<img width="876" alt="Screen Shot 2022-07-29 at 9 24 51 AM" src="https://user-images.githubusercontent.com/52292902/181781176-1231d7d9-e19c-4b5e-bc2f-8ded85be2518.png">

## What's the resolution?
<img width="1129" alt="Screen Shot 2022-07-29 at 9 26 00 AM" src="https://user-images.githubusercontent.com/52292902/181781377-de46b2fc-10d2-42fa-ba60-0852bbbcdf96.png">
Probably disable this `color` CSS property to get back white text?

<img width="777" alt="Screen Shot 2022-07-29 at 9 26 23 AM" src="https://user-images.githubusercontent.com/52292902/181781458-2ad43463-3342-4559-b75e-7fa860da3ede.png">

PR #92, opened 2022-08-01
fix darkmode table text

nhayfield:
Fixes https://github.com/pomerium/documentation/issues/87


desimone (approved):

desimone (changes requested):
Can we fix the action (up / down / more info) buttons too?
Screen Shot 2022-08-01 at 9 44 37 AM

desimone (approved):

PR #107, opened 2022-08-19
V19

nhayfield:
Fixes https://github.com/pomerium/internal/issues/961


alexfornuto:
@nhayfield I think you changed the wrong field. You made the source branch `v19` and the target branch remained `main`.

PR #159, opened 2022-11-11
Remove the reference to vals-entrypoint

i1snow:
the cloud run image in the latest (v0.19.1) does not use vals-entrypoint and the configuration through `VALS_FILES` does not work. The reference to the entrypoint needs to be removed and the instruction to add the config in cloud-run needs to be updated.


ZPain8464:
Hi @i1snow, thanks for the PR! I'm cc'ing @desimone  here to review your request.

desimone:
Tracking larger effort here. 
- https://github.com/pomerium/pomerium/issues/4091

PR #160, opened 2022-11-14
use ingress-controller/docs for k8s reference

wasaga:
`docs/k8s/reference.md` would be generated and pushed automatically whenever its source is updated in the `pomerium/ingress-controller` repo. 

docs generation in https://github.com/pomerium/ingress-controller/pull/428


content/docs/k8s/reference.md
```diff
@@ -368,7 +382,7 @@ IdentityProvider see https://www.pomerium.com/docs/identity-providers/
                 </p>
                 <p>
                     
-                    Scopes correspond to access <a href="https://www.pomerium.com/reference/#identity-provider-scopes">privilege scopes</a>.
+                    Scopes is a list of <a href="https://www.pomerium.com/reference/#identity-provider-scopes">privilege scopes</a> to request access to from the Identity Provider.
```
ZPain8464, line 385: This one still redirects to Reference homepage. Remove the "#" from the URL and it takes you to the correct page.

ZPain8464 (commented):
I found one link that still redirects to the reference homepage. Pending that change, this looks good!

desimone (approved):

PR #200, opened 2022-12-12
Fix links on routes page

alexrudd2:
The [routes page](https://www.pomerium.com/docs/concepts/routes) has two broken links:
- An attempted link to Reference: Routes goes nowhere.
- the link to Namepaces just goes back to the same page with a non-existent anchor

I'm not sure how these anchor-thingies work and have no way of testing it, but I think this PR fixed both.


desimone (approved):
Thank you @alexrudd2

ZPain8464 (approved):

PR #227, opened 2022-12-22
docs: redirects 404ing blog urls

ZPain8464:
Fixes: 
- https://github.com/pomerium/internal/issues/1127
- https://github.com/pomerium/internal/issues/1178
- https://github.com/pomerium/internal/issues/1180
- https://github.com/pomerium/internal/issues/1182
- https://github.com/pomerium/internal/issues/1177

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -70,8 +70,8 @@
 /posts/2020/04/16/release-0-7/ /blog/announcing-pomerium-0-7/
 
 # Redirects /blog/tag urls to /blog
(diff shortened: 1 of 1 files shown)


ZPain8464:
These still 404 based on the Netlify preview. So `blog/tag/istio` redirects to `blog/istio`, which isn't a tag at all. Should probably redirect these explicitly to `/blog`

desimone (approved):

PR #236, opened 2022-12-28
tcp: update advanced docs for new URL format

calebdoxsey:
Fixes https://github.com/pomerium/internal/issues/1203

You can now specify TCP routes with a separate path to indicate a specialized bastion host:

```
tcp+https://proxy.example.com:8443/ssh.example.com:22
```

This does the same thing the custom `pomerium-url` option did, but does so more explicitly so that it works properly with Pomerium TLS certificate selection.


cmo-pomerium (approved):
LGTM

Issue #216, opened 2022-12-20
update signing_key docs


calebdoxsey:
For https://github.com/pomerium/pomerium/pull/3828

We should update the reference documentation for `signing_key` and `signing_key_file` to note that they accept multiple keys.

PR #244, opened 2023-01-03
reference: update signing key docs

calebdoxsey:
Update signing key docs to mention multiple key support and add a section about key rotation.

Fixes https://github.com/pomerium/documentation/issues/216


ZPain8464 (approved):
I added some minor tweaks to the text to make the content more consistent. Otherwise, LGTM. Thanks for knocking this out, Caleb!

PR #271, opened 2023-01-11
update codec_type

calebdoxsey:
Remove the warning about codec type, which is no longer applicable because we now add all routes to all virtual hosts, so the http/2 coalescing problem doesn't bite us anymore.


ZPain8464 (approved):
Thanks for updating this

cmo-pomerium (commented):
didnt see anything that required my attention

Issue #273, opened 2023-01-11
Docs: Describe Kubernetes Ingress Controller settings


ZPain8464:
## What's incorrect or missing

Currently, we don't have reference documentation that details which Ingress Controller settings we don't support. 

## What's the resolution?

We need a list of "main reference" documentation to describe Ingress Controller keys, and cross-link relevant main reference docs to Ingress Controller reference docs.

@wasaga should also document a list of inclusion/exclusion criteria for Ingress settings.

PR #305, opened 2023-01-12
updated global reference with links to k8s reference

wasaga:
Fixes https://github.com/pomerium/documentation/issues/273


desimone (changes requested):
s/Kubernetes/Kubernetes Ingress Controller/g with a link pointing back to the docs on the ingress so it's clear this is for the ingress controller (and... say not helm)

desimone (approved):

desimone:
Links make it clear enough.

PR #343, opened 2023-01-24
Docs: moves enterprise.md file path

ZPain8464:
Moves `enterprise.md` to `/docs/releases/enterprise/install/enterprise`. 

Also fixes a few Enterprise links in Guides so they are relative instead of absolute.


desimone (approved):
Do we also need to update the redirects here? 

Otherwise lgtm.

PR #352, opened 2023-01-27
Docs:  split mode updates

ZPain8464:
From what I could tell, there weren't any instances that suggest a shared key or databroker params as a requirement for split service mode. I added a tip to the Enterprise Quickstart and a few other pages.


wasaga (approved):

wasaga (commented):
for shared_secret & databroker we need test it first; current wording in the pr is misleading, as discussed please close it and open a ticket with request to test and provide clarification wrt which parameters are now required. 

I think the only place it should be reflected is the reference guide.

wasaga (changes requested):
see above comment

PR #368, opened 2023-02-02
Docs: updates request flow diagram

ZPain8464:
I honestly think the sequence diagram is the best approach to explain interactions between components. I revamped the current one so it's easier to read (I added color-coded services and more context). I also include 3 steps in the diagram that are explained beneath it, but they are somewhat repetitive if you read the Components section as well. 

Another thing worth noting is the "miro" watermark in the bottom-right corner. I can update to a free trial to see if I can save the image in higher resolution without the watermark for the time being. If everyone is OK with this new diagram, I'll upgrade and replace the current image. 

Tagging @desimone, @calebdoxsey, and @wasaga for visibility.


calebdoxsey (approved):
LGTM

wasaga (approved):

content/docs/internals/architecture.md
```diff
@@ -1,4 +1,5 @@
 ---
+# cspell:ignore Authorization
```
desimone, line 2:
```suggestion
# cspell:ignore Authorization
```

Probably should just add tthis to `cspell.json` instead?

content/docs/internals/architecture.md
```diff
@@ -50,10 +51,30 @@ In test deployments, all four components may run from a [single binary and confi
...
-## Authentication Flow
+## The lifecycle of a request
 
-Pomerium's internal and external component interactions during full authentication from a fresh user are diagramed below.
+The diagram below shows how Pomerium's components communicate when authenticating a user.
 
 After initial authentication to provide a session token, only the authorization check interactions occur.
 
-![pomerium architecture diagram](./img/architecture/pomerium-auth-flow-stateless-auth.svg)
+![pomerium architecture diagram](./img/architecture/pomerium-request-flow-2.jpg)
```
desimone, line 60:
```suggestion
![pomerium architecture diagram](./img/architecture/pomerium-request-flow-2.jpg)
```

SVG would probably be better if you can export it.

content/docs/internals/architecture.md
```diff
@@ -50,10 +51,30 @@ In test deployments, all four components may run from a [single binary and confi
...
+
+**Step 1:** Unauthenticated client
+
+The client sends a request to access a secured application, which prompts the Proxy to send the request over a gRPC call to the Authorization service.
+
+Because the client has not authenticated, the Authorization service can't locate the session ID in the Databroker service. It sends the client to the Authenticate service, which prompts the client to sign in to the IdP.
+
+**Step 2:** Authenticating the client
+
+After the client signs in, the IdP exchanges an authorization code with the Authentication service. The Authentication service uses the code to get OAuth tokens and OIDC claims from the IdP.
```
desimone, line 70:
```suggestion
After the client signs in, the IdP exchanges an authorization code with the Authentication service. The Authentication service uses the code to get OAuth tokens and OIDC claims from the IdP.
```

> authorization code
we should probably link to the oidc spec which describes code flow.

desimone (changes requested):
There are two flow diagrams but only one is used. Also, SVG or a high quality png would be better than jpeg i think.

desimone (approved):

PR #408, opened 2023-02-16
dont require signing_key generated as it's already in the docker-compose

wasaga:


cmo-pomerium (approved):
approved at Denis' request

PR #416, opened 2023-02-21
Adding Pomerium 5 min vid

cmo-pomerium:
Simple add, but please check if there's an issue.

### DIFF
--- a/content/docs/index.mdx
+++ b/content/docs/index.mdx
@@ -47,3 +47,16 @@ Still feel too abstract? Hate text? Check out a two minute explainer demo of Pom
     allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
     allowFullScreen></iframe>
 </center>
+
+## Setup Pomerium in under 5 minutes
+
+<center>
+  <iframe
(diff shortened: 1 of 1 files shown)


ZPain8464 (changes requested):
So the video isn't "connecting" based on the deploy preview. Also, we should remove the 2-minute example video above if we include this one -- stacking two iframes is jarring. Thoughts?
SCR-20230221-cjl

content/docs/index.mdx
```diff
@@ -48,15 +48,9 @@ Still feel too abstract? Hate text? Check out a two minute explainer demo of Pom
     allowFullScreen></iframe>
 </center>
 
+
+
 ## Setup Pomerium in under 5 minutes
```
ZPain8464, line 53:
```suggestion
## Set up Pomerium in under 5 minutes
```

ZPain8464:
@cmo-pomerium you can get the pre-check to pass if you run `yarn format` before pushing your commit. Or, view the suggested change from prettier and manually make the change if you'd like.

ZPain8464 (approved):

PR #417, opened 2023-02-21
Update ppl.mdx

SakaZhang:
the example may be misleading


ZPain8464:
@SakaZhang thanks for the suggestion. Ccing @desimone for his input. I think extending the example is fine, thoughts?

desimone (approved):

PR #442, opened 2023-03-02
Docs: adds Jenkins guide

ZPain8464:
Includes Jenkins guide and images. 

If time permits, I'd love it if @ssveta7ak could test the Core and Enterprise paths before deploying.


ZPain8464:
@desimone I have the fix for the Jenkins guide, I just have to update a step in the Enterprise part. I'll get that finished tomorrow morning so we can push it passed the finish line.

ssveta7ak (approved):

PR #443, opened 2023-03-02
Docs: small update to device identity

ZPain8464:
Adds front matter and resources for further reading to Device Identity page.


cmo-pomerium (approved):
Approved at Zach's request

PR #462, opened 2023-03-07
Docs: cleans up redirects

ZPain8464:
This is just a housekeeping PR to clean up the redirects file:
- Deletes duplicate links 
- Rearranges links (somewhat) into categories.
- Labels links that use 2 or more lines (multiple redirects per link)
- Labels links that still 404 or redirect incorrectly (small handful)
I also found opportunities to consolidate links with a `:splat` but I'd like to do those in separate PRs one at a time so I can  test these out and make issues for them.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -93,6 +93,7 @@
     "proxying",
     "psql",
     "quickstart",
+    "quickstarts",
     "rego",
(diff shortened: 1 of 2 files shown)


desimone (approved):
f🤞

PR #490, opened 2023-03-10
Docs: redirects deprecated policy reference links

ZPain8464:
@desimone I'm a bit hesitant on this one because the last version supporting policy was v14. The URL structure uses a fragment identifier if I'm not mistaken (e.g. `/reference/#policy`).

We could modify the redirect so that it just goes to `/reference` and the user would have to scroll to find the Policy section. 
I also can't include the `#` in the `:splat` because it actually comments out the link, so linking to `reference` might be the only option.

Thoughts?


desimone (approved):

ZPain8464:
@wasaga I requested your review, even this was previously approved. I'd love it if you could review the kubernetes-access and ingress guide updates re: removing deprecated policy links.

content/docs/capabilities/kubernetes-access.md
```diff
@@ -42,7 +42,7 @@ Pomerium can be leveraged as a proxy for user requests to the API Server.
 Building on top of a standard Kubernetes and Pomerium deployment:
 
 1. Pomerium is given access to a Kubernetes service account with [impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation) permissions
-2. A [policy route](/docs/reference/policy/policy) is created for the API server and [configured](/docs/reference/routes/kubernetes-service-account-token) to use the service account token
+2. A [policy route](/docs/reference/routes/policy) is created for the API server and [configured](/docs/reference/routes/kubernetes-service-account-token) to use the service account token
```
wasaga, line 45: shouldn't it be the other way around - _the route's policy_ ?

content/docs/deploying/k8s/ingress.md
```diff
@@ -193,38 +190,6 @@ ingress.pomerium.io/allow_public_unauthenticated_access: 'true'
...
-**Array of strings**. Users with email domains matched would be allowed.
-
-</td></tr>
-<tr><td>
-
-```yaml
-ingress.pomerium.io/allowed_users: |
-  - alice@pomerium.com
-  - bob@contractor.co
-```
```
wasaga, line 210: could you please add an example below that would serve as a replacement, similar to the example with domains?

wasaga (approved):

ZPain8464:
@wasaga I added examples to the annotations. Could you please review real quick when you have a second?

wasaga (approved):

PR #553, opened 2023-03-21
Docs: restructures sidebar items

ZPain8464:
This restructure revolves primarily around these 4 sidebar items:
- Pomerium Core
- Enterprise
- Kubernetes
- Deploying

I left off the backport label to v21; if we decide to merge, it would require some redirects and might be jarring with the current layout. I'm not opposed to backporting, though. 

I'd also like to get the team's feedback on this restructure.


ZPain8464:
@desimone this is the layout, it still requires some changes to flesh out a few new pages and to correct the redirects file. But for now, you can take a look and see if there's anything you'd want to change or we can scrap the PR as well.

sidebars.js
```diff
@@ -6,15 +6,66 @@ const sidebars = {
...
+        },
+        'docs/enterprise/configure',
+        'docs/enterprise/upgrading',
+        'docs/enterprise/changelog',
+      ],
+      // items: [{type: 'autogenerated', dirName: 'docs/enterprise'}],
+    },
+    {
+      type: 'category',
+      label: 'Kubernetes',
```
desimone, line 47: Feels like this should live under "Deploy" thoughts?

sidebars.js
```diff
@@ -6,15 +6,66 @@ const sidebars = {
...
+          ],
+        },
+        'docs/pomerium-core/upgrading',
+        'docs/pomerium-core/changelog',
+      ],
+    },
+    {
+      type: 'category',
+      label: 'Enterprise',
+      link: {type: 'doc', id: 'docs/enterprise'},
```
desimone, line 26: And this should be under Enterprise?

desimone (commented):
Just took a look at the sidebar for now. Left two thoughts.

\cc @wasaga  who had some ideas here too so flagging

sidebars.js
ZPain8464: I moved "Kubernetes" because it has lots of pages, which increases the click depth to get to the right page. I was trying to avoid too many nested dropdowns, but it also adds an additional category to the sidebar. Can totally move this one under "Deploy", but also curious what @wasaga thinks?

PR #564, opened 2023-03-28
Docs: adds console changelogs

ZPain8464:
Adds:
- v0.20.1
- v0.21.0
- v0.21.1


desimone (commented):
Thanks @ZPain8464 ! Ideally, I'd like to simplify in keeping with our other releases for enterprise. Less commit history, more apple store changelog. Thoughts?

PR #570, opened 2023-04-07
Docs: updates core quickstart to run with hosted authenticate url

ZPain8464:


desimone (changes requested):
If we are cutting hosted authenticate, we can remove the idp_* settings here too!

PR #585, opened 2023-04-26
Docs: adds wildcard from routes config reference

ZPain8464:
Adds a Wildcard From Routes reference page. However, if this is Enterprise only it may make more sense to put this in the GUI Configuration page until we migrate all Console references/Settings to the main Reference page.


content/docs/reference/routes/from.md
```diff
@@ -10,16 +10,74 @@ pagination_next: null
...
 
 `From` is the externally accessible URL for the proxied request.
 
 Specifying `tcp+https` for the scheme enables [TCP proxying](/docs/capabilities/tcp/) support for the route. You may map more than one port through the same hostname by specifying a different `:port` in the URL.
 
+## Wildcard From Routes
+
+**Kubernetes:** Not supported <br/>
+
+**Wildcard From Routes** supports the use of a wildcard asterisk (`*`) prefixed to the domain name portion of a `from` URL.
```
calebdoxsey, line 43: it can occur anywhere within the domain name portion. It doesn't have to be a prefix.

calebdoxsey (changes requested):

calebdoxsey (approved):

content/docs/reference/routes/from.md
```diff
@@ -10,16 +10,72 @@ pagination_next: null
...
+# Or
+
+routes:
+  - from: tcp+https://*.example.com:22
+    to: tcp://example.com:22
+```
+
+:::tip **Note**
+
+Wildcard From Routes (`*`) may take precedence over non-wildcard routes regardless of their ordering. (To be safe, specify the ordering by using Wildcard From Routes after non-wildcard routes.)
```
desimone, line 75: We should mention that precedence of routes of more specified routes supersedes that of less specific wildcard routes.

content/docs/reference/routes/from.md
```diff
@@ -10,16 +10,72 @@ pagination_next: null
...
+
+---
 
 `From` is the externally accessible URL for the proxied request.
 
 Specifying `tcp+https` for the scheme enables [TCP proxying](/docs/capabilities/tcp/) support for the route. You may map more than one port through the same hostname by specifying a different `:port` in the URL.
 
+## Wildcard From Routes
+
+**Kubernetes:** Not supported <br/>
```
desimone, line 39: This is unofficially supported but the behavior of the two implementations are different. The behavior will be different than what is defined in the kubernetes docs. We should warn users that the behavior may be undefined. 

\cc @wasaga to describe distinction.

desimone (approved):

PR #634, opened 2023-05-16
Docs: add golangci-lint as a build prerequisite

kenjenkins:
On the Deploying / From Source page, add 'golangci-lint' to the Prerequisites list (linking to the local installation instructions). This is required by the `all` make target.


cmo-pomerium (approved):
LGTM

PR #635, opened 2023-05-17
Docs: adds signout-redirect-url ref to global config options

### LINKED ISSUE #633: signout_redirect_url is not a per-route option
Page: https://www.pomerium.com/docs/reference/routes/signout-redirect-url

This setting incorrectly resides under `routes`, while in fact it is not a per-route but a global option.

ZPain8464:
Fixes #633


wasaga:
Should it update redirects file?

wasaga (approved):

desimone (approved):

Issue #612, opened 2023-05-05
appleid docs are missing


desimone:
RELATED: https://github.com/pomerium/pomerium/pull/3959

## What's incorrect or missing

We don't have docs covering how to use this identity provider despite support for it.

## What's the resolution?

Write an identity provider doc similar to the others. 

https://docs.pomerium.com/docs/identity-providers

PR #641, opened 2023-05-24
Docs: adds apple idp guide

ZPain8464:
The Apple IdP is not configurable in the Console yet, so this guide only covers configuration for Core. I can create a follow-up PR if adding to Console is a priority in the near future.


ZPain8464:
Fixes #612

desimone (approved):
LGTM!

Only suggestion is we carry a similar callout that the providers may change like we do with the other guides

> CAUTION
> While we do our best to keep our documentation up to date, ......

PR #642, opened 2023-05-24
Docs: Updates Reference page

ZPain8464:
This PR mainly gets the ball rolling for combining Console, Core, and K8s references. Kubernetes (especially examples) will need a close review. Updates to the formatting should go a lot quicker as changes are requested. 

Two settings I noticed in `/docs/releases/enterprise/configure` – `grpc server max connection age` and `grpc server max connection age grace` – do not have references. Bringing these up in case we need to create refs for them.

### DIFF
--- a/content/docs/deploy/k8s/reference.md
+++ b/content/docs/deploy/k8s/reference.md
@@ -774,8 +774,6 @@ Storage defines persistent storage for sessions and other data. See <a href="htt
     </tbody>
 </table>
 
-
-
 ### `timeouts`
 
 Timeout specifies the <a href="https://www.pomerium.com/docs/reference/global-timeouts">global timeouts</a> for all routes.
@@ -844,8 +842,6 @@ Timeout specifies the <a href="https://www.pomerium.com/docs/reference/global-ti
     
     </tbody>
 </table>
-
-
 
 ## Status
 
--- a/content/docs/reference/branding/branding.md
+++ b/content/docs/reference/branding/branding.md
@@ -145,8 +145,8 @@ See [Custom Branding / Errors](/docs/capabilities/branding) for more information
 
 Customize the **Error Message Header** in the Console:
 
-  1. Enter your custom error message ![Error message box](./img/branding-error-message-header.png)
+1. Enter your custom error message ![Error message box](./img/branding-error-message-header.png)
 
-  1. Select **Show Error Details** for the route where you want to display the custom error message ![Show error details for a specific route](./img/branding-show-error-details.png)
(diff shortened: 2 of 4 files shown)


content/docs/reference/cookie-domain.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+| **Config file keys** | **Environment variables** |
+| :------------------- | :------------------------ |
+| `cookie_domain`      | `COOKIE_DOMAIN`           |
+
+</TabItem>
+<TabItem value="Enterprise" label="Enterprise">
+
+| **Config file keys** | **Environment variables** |
+| :------------------- | :------------------------ |
```
desimone, line 41: This looks wrong.

content/docs/reference/autocert/autocert-ca.mdx
```diff
@@ -0,0 +1,64 @@
...
+| **Config file keys** | **Environment variables** |
+| :------------------- | :------------------------ |
+| `autocert_ca`        | `AUTOCERT_CA`             |
+
+</TabItem>
+<TabItem value="Enterprise" label="Enterprise">
+
+| **Config file keys** | **Environment variables** |
+| :------------------- | :------------------------ |
+| `autocert_ca`        | `AUTOCERT_CA`             |
```
desimone, line 46: Looks wrong

desimone (changes requested):
Several of the pages have a tab for enterprise, but do not have enterprise items. If there is no enterprise setting, we should say so like we do with kubernetes.  For example, nearly everywhere in the enterprise tab where there is a key and not a photo or a video, the entries are copied from Core / is incorrect. For example:

- https://deploy-preview-642--pomerium-docs.netlify.app/docs/reference/cookie-secret
- https://deploy-preview-642--pomerium-docs.netlify.app/docs/reference/authenticate-internal-service-url
- https://deploy-preview-642--pomerium-docs.netlify.app/docs/reference/codec-type
- and way too many others to list

In general, it's difficult to review the code diffs because most of the renamed files are showing as deleted / adds instead of move / diffs (see signing-key.mdx for an example of where it's preserved). 

Otherwise, LGTM.

content/docs/reference/autocert/autocert-ca.mdx
ZPain8464: @desimone so this is what happens when you run `yarn format` (prettier) -- it adds dashes so that each row/column has the same amount of characters.

ZPain8464:
@desimone a handful of consolidated docs were included (IdP, branding, cookies). I slightly reformatted these single-page refs, but I think it's an improvement from what I was doing.

desimone (approved):

PR #644, opened 2023-05-26
Docs: sidebar restructure

ZPain8464:
This PR replaces `/docs/releases` with `/docs/deploy`. The **Deploy** sidebar houses the following sections:
- Pomerium Core
- Kubernetes
- Enterprise


ZPain8464:
The `redirects` file would need to updated as well. But I think it might be easier to merge this (if everyone is happy with it so far) and fix breaking links as they appear. (Similar to the last big re-org).

desimone (changes requested):
Overall, I think this is a good change. 

- I think we should move “Clients” from under “Pomerium core” to being under “Deploy” at the bottom of the list. 
- Move “To production” from Enterprise to under Deploy and move it to the very bottom of the list. 


Otherwise, LGTM pending … yeah, that’s going to be a lot of redirects. :/

static/_redirects
```diff
@@ -336,7 +336,7 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 # flattened the changelog, and upgrading guide
 /docs/overview/upgrading/archive  /docs/overview/upgrading
 /docs/overview/changelog/archive  /docs/overview/upgrading
-/docs/overview/upgrading /docs/releases/upgrading
+/docs/overview/upgrading /docs/deploy/core/upgrading
```
desimone, line 339: Should we set a catch all for `/deploying/*` and `/releases/*`?

desimone (approved):

PR #682, opened 2023-06-08
Docs: cleans up external data enterprise config

ZPain8464:
This PR moves the External Data content from `/docs/deploy/enterprise/configure` to `/docs/integrations`. 

It also adds a GIF of the external settings page.


desimone (approved):
Gif looks nice.

PR #710, opened 2023-06-13
Docs: updates set_request_headers reference

ZPain8464:
Fixes https://github.com/pomerium/pomerium/issues/4214

### DIFF
--- a/content/docs/reference/routes/set-authorization-header.mdx
+++ b/content/docs/reference/routes/set-authorization-header.mdx
@@ -15,6 +15,14 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
+:::caution
+
+As of v0.24.0, `set_authorization_header` has been **deprecated** in favor of [`set_request_headers`](/docs/reference/routes/set-request-headers) and will be removed in future versions.
+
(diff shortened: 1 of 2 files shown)


nhayfield (approved):

content/docs/reference/routes/set-authorization-header.mdx
```diff
@@ -15,6 +15,14 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
+:::caution
+
+As of v0.24.0, `set_authorization_header` has been **deprecated** in favor of [`set_request_headers`](/docs/reference/routes/set-request-headers) and will be removed in future versions.
```
kenjenkins, line 20: I'm confused about the version timeline here.

My impression was:
 - v0.22 is the current release
 - in v0.23 (the next release) set_authorization_header is deprecated
 - in v0.24 (the release after that) set_authorization_header will be removed

Is that not the plan?

ZPain8464: @ZPain8464 you're correct, and thanks for calling that out. I can follow up this PR with another to correct the version number mentioned here.

PR #718, opened 2023-06-15
Docs: corrects `set_authorization_header` version number and updates text

ZPain8464:
This PR updates the incorrect version numbers listed in #710. It also rewrites the `set_request_headers` Summary. 

Tagging @kenjenkins and @ssveta7ak here.

### DIFF
--- a/content/docs/reference/routes/set-request-headers.mdx
+++ b/content/docs/reference/routes/set-request-headers.mdx
@@ -18,22 +18,26 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
-**Set Request Headers** allows you to set both static and dynamic values in request headers for downstream applications.
+**Set Request Headers** allows you to set both static and dynamic values for given request headers. Static values can be useful if you want to pass along additional information to upstream applications as headers, or to set a fixed authentication header on the request.
 
-You can use static header values to pass additional information to downstream applications, or to send static credentials in an Authorization header (see [Basic Authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#basic_authentication) for more information).
+The dynamic values enable you to pass ID and Access tokens from your identity provider to upstream applications.
 
-You can use dynamic header values to pass ID and Access tokens in your requests to downstream applications.
+To pass dynamic values from the user's OIDC claim to an upstream service, see [JWT Claim Headers](../jwt-claim-headers).
 
-See the configuration options below for examples of static and dynamic headers applied to a route.
+:::caution
+
+Neither HTTP/2 [psuedo-headers](https://www.rfc-editor.org/rfc/rfc9113.html#PseudoHeaderFields) (e.g. `:authority`) nor the `Host:` header may be modified via this mechanism. Those headers may instead be modified via [`prefix_rewrite`](/docs/reference/routes/prefix-rewrite), [`regex_rewrite`](/docs/reference/routes/regex-rewrite), and [`host_rewrite`](/docs/reference/routes/host-rewrite).
+
+:::
 
 ## How to configure
 
 <Tabs>
 <TabItem value="Core" label="Core">
 
-| **YAML**/**JSON** setting | **Type** | **Usage**    |
-| :------------------------ | :------- | :----------- |
-| `set_request_headers`     | `string` | **optional** |
+| **YAML**/**JSON** setting | **Type**                          | **Usage**    |
+| :------------------------ | :-------------------------------- | :----------- |
+| `set_request_headers`     | `string` (map of key-value pairs) | **optional** |
 
 </TabItem>
 <TabItem value="Enterprise" label="Enterprise">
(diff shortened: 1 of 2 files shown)


content/docs/reference/routes/set-request-headers.mdx
```diff
@@ -52,13 +54,15 @@ See Kubernetes [Ingress](/docs/deploy/k8s/ingress#annotations) for more informat
...
-### Pass ID and Access tokens
+### Pass dynamic values in headers
 
 | **Option** | **Value** | **Type** | **Usage** |
 | :-- | :-- | :-- | :-- |
 | `x-pomerium-downstream-id-token` | `$pomerium.id_token` | Map of key-value pairs | **optional** |
 | `x-pomerium-downstream-access-token` | `$pomerium.access_token` | Map of key-value pairs | **optional** |
 
+**Note:** Token values must use the `$pomerium.<token>` syntax.
+
```
kenjenkins, line 65: This table (and the one above it) doesn't make sense to me. I don't think the "Option" heading is accurate, and I think the "Map of key-value pairs" belongs in the previous table, under the "Type" heading.

If you like I'd be happy to update this section.

kenjenkins (commented):
The version number change looks fine to me, but I have some concerns about the rest of this page.

content/docs/reference/routes/set-request-headers.mdx
ZPain8464: @kenjenkins the first table lists the configuration key, data type, and whether the setting is required/optional. It's the original setting to be used with static header values (I reformatted that data into a table, but before it was an unordered list). 

The second table addresses the new functionality that allows you to pass in dynamic values into headers for the tokens. We've used `map of key-value pairs` in other config settings, so I've applied it here.  As for labeling this table **Options**, I'm open to calling it something else. 

What about it doesn't make sense? Asking because we've applied the same formatting to pretty much every reference entry so far.

I'm hesitant to hand it off entirely just because I want to maintain consistency with our reference entries, but I'd like to see what your suggestion is.

kenjenkins: Here's a concrete proposal, hopefully that can help clarify my concerns: https://github.com/pomerium/documentation/commit/7a54d1d83f91a9e86f79304a0f451e9452e0ff27

Specifically:
 - the **Type** column in the first table does not make sense to me — as I understand it this option is not a `string`, it is a mapping (both before and after the change to support the new `$pomerium.<token>` values)
 - the available token options (`$pomerium.id_token` and `$pomerium.access_token`) are not themselves key-value pairs; rather they can be used as part of the value
 - it doesn't make sense to me to include example header names (`x-pomerium-downstream-id-token`) in the table, as these are up to the user to decide (right?)

Does that make sense?

ZPain8464: Hey Ken, so:
- That's fine, we can update that to `map of key-value pairs` (or the suggestion in your proposal above)
- As for the token options (`$pomerium.<token>`), I'm aware those aren't key-value pairs, the confusion lies in the point below
- I was under the assumption that those header values (what I thought of as the `key`) weren't customizable, but hardcoded values (`x-pomerium-downstream-id-token`), hence the reason I included them in a table as an Option, and the value as `$pomerium.<token>`. 

Thanks for taking the time to check this out to make sure it's technically accurate. What I'll do is update this PR with your suggestions (early tomorrow). I'll ping you on Slack so you can review.

content/docs/reference/routes/set-request-headers.mdx
```diff
@@ -18,20 +19,26 @@ import TabItem from '@theme/TabItem';
...
 
 <Tabs>
 <TabItem value="Core" label="Core">
 
-| **YAML**/**JSON** setting | **Type** | **Usage**    |
-| :------------------------ | :------- | :----------- |
-| `set_request_headers`     | `string` | **optional** |
+| **YAML**/**JSON** setting | **Type**                          | **Usage**    |
+| :------------------------ | :-------------------------------- | :----------- |
+| `set_request_headers`     | `string` (map of key-value pairs) | **optional** |
```
kenjenkins, line 41: I still think it's a little confusing to start with `string` here. If you want to use "map of key-value pairs" I'd suggest removing the `string` part entirely.

I also wonder if the Kubernetes table **Type** column should be updated too? @wasaga: is the Kubernetes property also a map, or is it encoded as a plain string somehow?

kenjenkins (approved):

content/docs/reference/routes/set-request-headers.mdx
```diff
@@ -52,16 +59,24 @@ See Kubernetes [Ingress](/docs/deploy/k8s/ingress#annotations) for more informat
...
+| **Token**                | **Value**                                     |
+| :----------------------- | :-------------------------------------------- |
+| `$pomerium.id_token`     | OIDC ID token from the identity provider      |
+| `$pomerium.access_token` | OAuth access token from the identity provider |
+
+**Note:** Token values must use the `$pomerium.<token>` syntax.
+
+:::warning
+
+Be very careful when passing access tokens to an upstream application. This may allow the application to make other authenticated requests on behalf of the user.
```
desimone, line 73: LGTM!

desimone (approved):

PR #767, opened 2023-06-26
Docs: adds idp polling delays troubleshooting

### LINKED ISSUE #717: Add IdP polling delays troubleshooting guide
We currently don't offer any guidance on how to adjust IdP polling delay settings. Several customers with sizable directories have had issues where the directory sync never completes. Adding a troubleshooting guide to resolve IdP polling delay issues (mentioned in this [Slack thread](https://pomerium-com.slack.com/archives/C0277SH5KL3/p1686796602365519))  would help larger orgs who may experience this issue in the future.

This troubleshooting guide/section ticket is linked to #716.

ZPain8464:
Resolves #717 with a small troubleshooting guide. The guide uses GIFs to guide the reader with written steps, and links to the IdP Polling reference.


content/docs/internals/troubleshooting.mdx
```diff
@@ -203,6 +203,26 @@ Usually, this is the result of either a routing issue or a configuration error.
 
 <GenerateRecoveryToken />
 
+### Directory sync
+
+**Problem:**
+
+When syncing your identity provider directory to the Enterprise Console, you may have issues completing the sync due to timeouts or errors.
```
kenjenkins, line 210: Would it make sense to phrase the problem more like this?

```suggestion
Enterprise Console directory sync fails to complete.
```

(I see the other troubleshooting sections aren't really set up like Problem / Solution, but I wonder if it would make sense to give a very brief "Problem" summary for the others too.)

kenjenkins (approved):
Nice!

Not related to this change specifically, but I wonder if we should make the Troubleshooting page its own top-level item in the sidebar? I'm not sure I would describe this as "Internals".

Also, if it were a top-level section, maybe it could have an overview page with just the "Problem" summaries, and links to individual pages for each problem scenario? (Just a thought.)

content/docs/internals/troubleshooting.mdx
ZPain8464: @kenjenkins yeah I figured I'd try this format, it seemed to fit this scenario well. Maybe we could clean up the existing sections and add this formatting in a separate PR? Could also take your idea to make this a top-level page into account in that PR.

kenjenkins: Definitely, a separate PR sounds good to me.

PR #773, opened 2023-06-27
Document idp_request_params default values

kenjenkins:
Add a 'Defaults' sub-section to the Identity Provider Request Params documentation, listing the default request parameters for Apple, Azure, and Google. Add a note about the newly implemented support for clearing the default parameters entirely (from pomerium/pomerium#4315).

Also list the supported identity provider types explicitly, under Identity Provider Name.

Add a custom CSS rule to avoid some unwanted extra space when placing a code block inside a table cell.

### DIFF
--- a/content/docs/reference/identity-provider-settings.mdx
+++ b/content/docs/reference/identity-provider-settings.mdx
@@ -157,6 +157,20 @@ IDP_CLIENT_SECRET_FILE='/run/secrets/POMERIUM_CLIENT_SECRET'
 
 See [identity provider](/docs/identity-providers/) for details.
 
+The supported values for this setting are:
+
+- `apple`
+- `auth0`
+- `azure`
+- `cognito`
+- `github`
+- `gitlab`
+- `google`
+- `oidc`
+- `okta`
+- `onelogin`
+- `ping`
+
 ### How to configure
 
 <Tabs>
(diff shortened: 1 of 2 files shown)


content/docs/reference/identity-provider-settings.mdx
```diff
@@ -195,7 +209,17 @@ IDP_PROVIDER=github
 
 **Identity Provider Request Params** lists the parameters you want to include as part of a sign-in request using the OAuth 2.0 code flow.
 
-Downstream application headers will be overwritten by Pomerium's headers on conflict.
```
kenjenkins, line 198: I don't understand what this means, and I'm wondering if we should remove it. (What's a "downstream application" in the context of an identity provider request?)

desimone: I _believe_ downstream _request_ headers will be overwritten on conflict? I guess I'm not sure either why it's relevant.

content/docs/reference/identity-provider-settings.mdx
```diff
@@ -195,7 +209,17 @@ IDP_PROVIDER=github
...
+
+Pomerium includes some default parameters for specific identity providers. Setting this configuration option will replace these default parameters. To remove the default parameters entirely, set this option to an empty map\* (e.g. `idp_request_params: {}` in the config file).
+
+| **Provider** | **Default parameters** |
+| :-- | :-- |
+| `apple`\* | <pre>response_mode: form_post</pre> |
+| `azure` | <pre>prompt: select_account</pre> |
+| `google` | <pre>prompt: select_account consent<br />access_type: offline</pre> |
+
+\*_Note:_ the default parameters for `apple` behave differently; any values set using this configuration option will be merged with the default parameters.
```
ZPain8464, line 222: (Just a suggestion)
```suggestion
:::note 

\*The default parameters for `apple` behave differently; any values set using this configuration option will be merged with the default parameters.

:::
```

content/docs/reference/identity-provider-settings.mdx
```diff
@@ -195,7 +209,17 @@ IDP_PROVIDER=github
 
 **Identity Provider Request Params** lists the parameters you want to include as part of a sign-in request using the OAuth 2.0 code flow.
 
-Downstream application headers will be overwritten by Pomerium's headers on conflict.
+### Defaults
```
ZPain8464, line 212: Not a huge deal, but could we move the **Defaults** section below **How to configure**? Just because that's the order we follow for the rest of these settings.

ZPain8464 (approved):

content/docs/reference/identity-provider-settings.mdx
kenjenkins: Thanks, fixed. I thought I was following the pattern from the other sections, but clearly I did not. :/

PR #806, opened 2023-07-07
update Go SDK example

kenjenkins:
Update the Go SDK example to account for API changes in the v0.0.9 Go SDK release.


ZPain8464 (approved):

ZPain8464:
@kenjenkins This doesn't need to be backported, right?

PR #808, opened 2023-07-09
Fix certificate example using wrong config key

brian-mcnamara:


desimone (approved):
Thanks for the fix @brian-mcnamara

PR #864, opened 2023-07-20
Docs: updates idp provider reference page

### LINKED ISSUE #863: `idp_provider` usage note may be incorrect
Page: https://www.pomerium.com/docs/reference/identity-provider-settings#identity-provider-name

'How to configure' shows the following table:

> Config file keys | Environment variables | Type | Usage
> -- | -- | -- | --
> idp_provider | IDP_PROVIDER | string | **required** (unless [self-hosting](https://www.pomerium.com/docs/capabilities/self-hosted-authenticate-service))

I think the "**required** (unless self-hosting)" may be incorrect. Isn't it the other way around? `idp_provider` is required only when self-hosting the authenticate service?

## What's the resolution?

Should this say "**required** (if [self-hosting](https://www.pomerium.com/docs/capabilities/self-hosted-authenticate-service))" instead? This would match the entry for [Identity Provider Client ID](https://www.pomerium.com/docs/reference/identity-provider-settings#identity-provider-client-id).

ZPain8464:
Fixes #863

### DIFF
--- a/content/docs/reference/identity-provider-settings.mdx
+++ b/content/docs/reference/identity-provider-settings.mdx
@@ -178,7 +178,7 @@ The supported values for this setting are:
 
 | **Config file keys** | **Environment variables** | **Type** | **Usage** |
 | :-- | :-- | :-- | :-- |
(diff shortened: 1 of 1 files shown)


kenjenkins (approved):
Thanks for the quick fix!

PR #922, opened 2023-08-17
mtls: note client_ca is deprecated 

kenjenkins:
Add a note that the `client_ca` setting is deprecated in favor of the new `downstream_mtls.ca` setting.

Remove the `client_crl` setting entirely, as this doesn't work in current Pomerium releases.


ssveta7ak (approved):

desimone (approved):
Did you want to split this into two PRS? One that can be backported to all the versions where `client-crl` didn't work, and another that would be merged to main/0-23-0 covering the deprecation notice?

PR #930, opened 2023-08-21
Update Quickstart to be a better UX

cmo-pomerium:
This is the beginning of a quickstart makeover. Zach will provide the technical documentation expertise while Colin is here to ensure that the experience feels more guided and more of a "now it begins!" than "this is the end"


cmo-pomerium:
@ZPain8464 I see that a lot of things failed, but the PR is for you to take over. The Quickstart changes are just what I thought would be a useful change for "next step!" and then we'll work on the guided tutorial, which should include the First Route

That being said, I think the First Route guide should be written as a standalone, and the one we make for the Guided Tutorial will  be custom written with the entire flow in mind.

PR #941, opened 2023-08-24
Docs: updates Core upgrade guide header formatting

ZPain8464:
I noticed, after incrementing each header, that we're missing v18 and v19 here. In the Console Upgrade guide, we're also missing v19. Should we add those to Core and Console? Not sure why those were omitted.


kenjenkins (approved):
I'll defer to @desimone about the v0.18 and v0.19 question.

desimone (approved):

PR #942, opened 2023-08-24
Docs: v23 core release

ZPain8464:
v23 Core Changelog and Upgrade guide.


content/docs/deploy/core/upgrading.mdx
```diff
@@ -13,6 +13,122 @@ sidebar_position: 10
 
 # Upgrade guide
 
+## Since 0.22.0
```
kenjenkins, line 16: This should be "0.23.0", right? (to be consistent with https://github.com/pomerium/documentation/pull/941)

```suggestion
## 0.23.0
```

kenjenkins (approved):

content/docs/deploy/core/upgrading.mdx
ZPain8464: Yes, but in the event that #941 doesn't get merged before the release, I figured I'd stick to the current pattern.

desimone (approved):

PR #943, opened 2023-08-24
Docs: adds v23 Console changelog and upgrade guides

ZPain8464:
Related to https://github.com/pomerium/internal/issues/1506

Adds Console Changelog and Upgrade guides for v23.


content/docs/deploy/enterprise/changelog.mdx
```diff
@@ -7,6 +7,18 @@ toc_max_heading_level: 5
 
 # Changelog
 
+## 0.23.0
+
+### New
+
+- [Set Request Headers](/docs/reference/routes/headers#set-request-headers) has a new token substitution value that sends the client certificate fingerprint (downstream mTLS must be enabled) to the upstream application or service.
```
kenjenkins, line 14: We should probably mention the ID token and access token substitutions here as well.

content/docs/deploy/enterprise/changelog.mdx
```diff
@@ -7,6 +7,18 @@ toc_max_heading_level: 5
...
+- [Access Log Fields](/docs/reference/access-log-fields) and [Authorize Log Fields](/docs/reference/authorize-log-fields) settings make it easier for you to review HTTP request logs from the Proxy and Authorize services.
```
kenjenkins, line 15: These settings let you customize the existing logs... maybe we could tweak this description a little to clarify that?

```suggestion
- [Access Log Fields](/docs/reference/access-log-fields) and [Authorize Log Fields](/docs/reference/authorize-log-fields) settings allow you to customize the values that are logged in the access and authorize logs.
```

content/docs/deploy/enterprise/upgrading.mdx
```diff
@@ -7,6 +7,34 @@ description: This page contains the list of deprecations and important or breaki
...
 
+## 0.23.0
+
+### Before you upgrade
+
+#### New
+
+**Set Request Headers**<br/>
+
+[Set Request Headers](/docs/reference/routes/headers#set-request-headers) has a new token substitution value that sends the client certificate fingerprint (downstream mTLS must be enabled) to the upstream application or service.
```
kenjenkins, line 18: Are new features typically mentioned in the upgrading guide? I wonder if it's sufficient just to mention these in the changelog.

kenjenkins (approved):

content/docs/deploy/enterprise/upgrading.mdx
ZPain8464: In the past, they've called out new features and also how to upgrade. I'm fine removing, but would like to see what @desimone thinks before we remove it.

content/docs/deploy/enterprise/changelog.mdx
desimone: It looks like we started co-mingling upgrade notes and new functionality around v0.17. Traditionally, upgrade guide should _just_ be what the user needs to know before they upgrade, and any steps they need to take. I think we should get back to that and leave text like this in the announcement. We can link to the announcement posts here however which might be nice.

content/docs/deploy/enterprise/upgrading.mdx
ZPain8464: @desimone I cleaned this up. Please let me know if the changes are OK!

desimone (approved):

Issue #866, opened 2023-07-20
Add "OIDC" under "Identity Providers" sidebar section?


kenjenkins:
## What's incorrect or missing

Today we had a user mention that they didn't see any support for a generic OIDC identity provider (see https://github.com/pomerium/ingress-controller/pull/697#issuecomment-1644097333). I believe we do have at least some support for generic OIDC, but we don't have a separate page for it under the "Identity Providers" section.

## What's the resolution?

Should we add an "OIDC" or "Generic OIDC" page under the "Identity Providers" section in the sidebar? Is there a specific identity provider we could add as an example there?

PR #973, opened 2023-09-06
Docs: adds keycloak OIDC guide

ZPain8464:
Resolves https://github.com/pomerium/internal/issues/1489. Also resolves https://github.com/pomerium/documentation/issues/866

This PR adds OIDC to the sidebar under `identity provider`. It adds an OIDC implementation using Keycloak as the OIDC server. 

This guide is only compatible with Core at the moment.


content/docs/identity-providers/oidc.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+## Set up Keycloak
+
+In your terminal, run Docker Compose:
+
+```yaml
+docker compose up
+```
+
+When Keycloak is ready, navigate to `http://0.0.0.0:8080` to access the **Administration Console:**
```
kenjenkins, line 79: I'm not sure that all browsers will translate `0.0.0.0` to `localhost`. I'd recommend we specify `localhost` explicitly here instead.

I think we can also make this into a link?

```suggestion
When Keycloak is ready, navigate to [http://localhost:8080](http://localhost:8080) to access the **Administration Console:**
```

(and similarly for other occurrences of 0.0.0.0 in this page)

content/docs/identity-providers/oidc.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+## Set up Pomerium
+
+Create a YAML file and add the following configuration:
+
+```yaml title="config.yaml"
+authenticate_service_url: https://authenticate.localhost.pomerium.io
+
+idp_provider: oidc
+idp_client_id: 'account'
```
kenjenkins, line 149: Shouldn't this use the client created in the previous step?

content/docs/identity-providers/oidc.mdx
```diff
@@ -0,0 +1,204 @@
...
+idp_provider_url: 'http://keycloak.localhost.pomerium.io:8080/realms/MyRealm'
+
+signing_key: <generate_key>
```
kenjenkins, line 152: I wonder if this will confuse some users. Would it be simpler to provide a valid key here instead, along with a warning not to use it in production?

content/docs/identity-providers/oidc.mdx
```diff
@@ -0,0 +1,204 @@
...
+## Set up Pomerium
+
+Create a YAML file and add the following configuration:
+
+```yaml title="config.yaml"
+authenticate_service_url: https://authenticate.localhost.pomerium.io
+
+idp_provider: oidc
+idp_client_id: 'account'
+idp_provider_url: 'http://keycloak.localhost.pomerium.io:8080/realms/MyRealm'
```
kenjenkins, line 150: I'd recommend we pick one realm name and use it consistently throughout this guide. It looks like the screen recording above uses the realm name 'Pomerium'.

We could update line 93 to read:
> In the **Realm name** field, enter the name "Pomerium"

and this line to read:
> `idp_provider_url: 'http://keycloak.localhost.pomerium.io:8080/realms/Pomerium'`

content/docs/identity-providers/oidc.mdx
```diff
@@ -0,0 +1,192 @@
...
+1. Keep **Client type** as **OpenID Connect**
+1. In the **Client ID** field, enter a name like `mynewclient`
+1. In the **Authentication flow** field, select **Standard flow** and **Direct access grants**
+1. Leave **Login settings** empty, and select **Save**
+
+![Create a client](./img/oidc/keycloak-create-client.gif)
+
+Update the client you created:
+
+1. Under **General Settings** in the **Name** field, enter your Client ID in template literals (for example, `${mynewclient}`) ![Enter client ID](./img/oidc/keycloak-client-general-settings.png)
```
kenjenkins, line 133: I don't understand the `${mynewclient}` part here. What does this do?

kenjenkins (approved):
Overall looks good to me.

content/docs/identity-providers/oidc.mdx
ZPain8464: It specifies the display name of the client. I tested it without this value and it works fine, but I was getting an invalid client credentials error today after testing again. I had to add the client secret to get it work. I added these steps as well.

PR #981, opened 2023-09-12
Docs: adds v18 and v19 notices to core and console upgrade guides

ZPain8464:
Fixes https://github.com/pomerium/internal/issues/1526


desimone (approved):
Added a bunch of backports; lmk if they make sense.

kenjenkins (approved):

ZPain8464:
@desimone I'd have to update the v20 - v22 Upgrade Guides formatting. In these releases, we're using the "Since 0.22.0" header scheme, but we updated v23 to be "0.23.0". I can push separate PRs to update these three versions, though.

PR #1019, opened 2023-10-02
fix helm guide formatting

### LINKED ISSUE #996: Some of the Helm guide content appears mangled
Page: https://www.pomerium.com/docs/guides/helm

It looks like some of the page markup displays within a code block:

Screen Shot 2023-09-19 at 10 22 35 AM

It looks like this may have been caused by some bad formatting changes introduced along with https://github.com/pomerium/documentation/commit/dd2a6a97ba964b70764392571d756a084d97d168.

## What's the resolution?

We could fix the formatting issues, but alternatively should we go ahead and delete the Helm guide altogether? If I understand correctly we no longer maintain a Helm chart for Pomerium. Should we remove this guide page at some point?

kenjenkins:
Some of the Helm guide Markdown was indented too far, causing it to be rendered as a code block. Update the formatting to restore this guide content.

Fixes #996.

### DIFF
--- a/content/docs/guides/helm.mdx
+++ b/content/docs/guides/helm.mdx
@@ -174,7 +174,7 @@ If you haven't already, install cert-manager and create a CA issuer. You can fol
     :::
 
     <details>
-    <summary>Default Certificate</summary>
+<summary>Default Certificate</summary>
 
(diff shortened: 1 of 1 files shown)


ZPain8464 (changes requested):
If we can find a way to format that's compatible with prettier, we can get the pre-commit check to pass.

ZPain8464 (approved):

Issue #1007, opened 2023-09-29
Footer newsletter sign-up form text has poor contrast in dark mode


kenjenkins:
Page: n/a

## What's incorrect or missing

The colors of the footer newsletter sign-up form do not change in dark mode, making the "Your email is safe with us" text nearly illegible.

<img width="341" alt="Screen Shot 2023-09-29 at 9 47 25 AM" src="https://github.com/pomerium/documentation/assets/51246568/35b265da-1336-4499-9f63-c665fb806cad">

## What's the resolution?

We should update this form to use different colors in dark mode.

PR #1024, opened 2023-10-02
Docs: updates footer ns color

ZPain8464:
I used the same var() function we use for Footer Links to toggle font color based on the dark/light theme. This seems to work well. 

Fixes https://github.com/pomerium/documentation/issues/1007


desimone:
Screenshot 2023-11-28 at 1 27 35 PM

src/css/custom.css
```diff
@@ -127,7 +127,7 @@
 }
 
 .text-slate\/60 {
-  color: rgb(67 84 101 / 0.6);
+  color: var(--ifm-footer-color);
```
desimone, line 130: What does this look like for you?

desimone (changes requested):

src/css/custom.css
ZPain8464: SCR-20231219-piak

ZPain8464: The button could use a `{cursor: pointer}` rule as well so the mouse switches from the arrow to the glove.

ZPain8464:
I updated it, now it looks like this: 
footer-contrast

src/css/custom.css
```diff
@@ -127,7 +127,7 @@
 }
 
 .text-slate\/60 {
```
kenjenkins, line 129: I think the rule name is a little odd: it feels like this was the name of a specific color. Should we consider renaming this rule if it's going to apply a theme color instead?

ZPain8464:
@kenjenkins i updated the rule name, but not sure it's any more effective than the last one. Open to suggestions, also curious what the best practice is here in terms of CSS naming conventions?

src/css/custom.css
```diff
@@ -126,7 +126,7 @@
   text-align: center;
 }
 
-.text-slate\/60 {
+.text-email\/60 {
```
kenjenkins, line 129: nit: let's drop the "\/60" part too

(My guess is that "slate/60" was a color name from some color system that we're not really using.)

kenjenkins (approved):

desimone (approved):

PR #1077, opened 2023-11-28
Docs: Clarifies route matching order

ZPain8464:
Resolves https://github.com/pomerium/pomerium/issues/4721

### DIFF
--- a/content/docs/capabilities/routing.mdx
+++ b/content/docs/capabilities/routing.mdx
@@ -409,20 +409,15 @@ routes:
 
 ## Route matching order
 
-If your configuration includes multiple routes with identical definitions, Pomerium will process these routes in the order they are defined in your configuration file.
+Pomerium processes routes in the order they are defined in its configuration. If you’re using Pomerium Core, then Pomerium will source its configuration from a configuration file.
 
-In the example below, if Pomerium receives a request to access `https://a.example.com:8080`, Pomerium will first check `https://a.example.com:8080/api` before matching the request with `https://a.example.com:8080`.
+In Enterprise Console and Ingress Controller deployments, routes are defined in Pomerium’s Databroker service. In this case, Pomerium sorts routes deterministically in the following order:
 
-```yaml title="config.yaml"
-- from: https://a.example.com:8080
-  to: http://example:8080
-  prefix: /api
-  allow_any_authenticated_user: true
+1. Ascending by `host`.
+1. Descending by `path`.
+1. Descending by `regex`.
+1. Descending by `prefix`.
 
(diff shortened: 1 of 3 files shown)


content/docs/capabilities/routing.mdx
```diff
@@ -407,6 +407,23 @@ routes:
       - http://myapp-srv-5:8080,10
 ```
 
+## Route matching order
+
+If your configuration includes multiple routes with identical definitions, Pomerium will process these routes in the order they are defined in your configuration file.
```
wasaga, line 412: pomerium configuration may be sourced from config file and databroker configuration. 

anything defined in the config file is processed ahead of whatever comes from the databroker, and yes it is processed in the order as it's listed in the config file. 

routes that are defined in the databroker, which is the case for the ingress controller, enterprise console and zero, are sorted in the deterministic manner: 

1. Ascending by `host`.
2. Descending by `path`.
3. Descending by `regex`.
4. Descending by `prefix`.

wasaga (changes requested):

content/docs/capabilities/routing.mdx
```diff
@@ -407,6 +407,17 @@ routes:
       - http://myapp-srv-5:8080,10
 ```
 
+## Route matching order
+
+Pomerium processes routes in the order they are defined in its configuration. If you’re using Pomerium Core, then Pomerium will source its configuration from a configuration file.
+
+In Enterprise Console and Ingress Controller deployments, routes are defined in Pomerium’s Databroker service. In this case, Pomerium sorts routes deterministically in the following order:
```
kenjenkins, line 414: I think it would be good to incorporate Denis's feedback about the matching order between config file and databroker, maybe something like this?

```suggestion
Routes are matched in a specific order; an incoming request will be sent to the first route that matches. Routes defined in a configuration file are considered first, in the order in which they appear in the configuration file, followed by any routes defined in the Enterprise Console or by the Ingress Controller. Routes from the Enterprise Console or Ingress Controller are sorted in the following order:
```

content/docs/capabilities/routing.mdx
```diff
@@ -407,6 +407,17 @@ routes:
       - http://myapp-srv-5:8080,10
 ```
 
+## Route matching order
+
+Pomerium processes routes in the order they are defined in its configuration. If you’re using Pomerium Core, then Pomerium will source its configuration from a configuration file.
```
kenjenkins, line 412: I wonder if it would also be good to include an example here where the order is important? e.g. a route with a path prefix and then another route for the same domain without any path prefix?

wasaga (approved):

PR #1109, opened 2023-12-14
Docs: updates k8s settings to use new name convention

ZPain8464:
@wasaga I've updated all the settings I could based on what we discussed. I have a list of settings we need to discuss that I'll fix in a follow-up PR. 

Related to https://github.com/pomerium/internal/issues/1630

### DIFF
--- a/content/docs/reference/routes/allow-any-authenticated-user.mdx
+++ b/content/docs/reference/routes/allow-any-authenticated-user.mdx
@@ -47,7 +47,7 @@ Enable **Any Authenticated User** in the **Policy Builder** in the Console:
 </TabItem>
 <TabItem value="Kubernetes" label="Kubernetes">
 
-| **Annotation name**                       | **Type**  | **Default** | **Usage**    |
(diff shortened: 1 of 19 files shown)


content/docs/reference/routes/allow-any-authenticated-user.mdx
```diff
@@ -47,7 +47,7 @@ Enable **Any Authenticated User** in the **Policy Builder** in the Console:
 </TabItem>
 <TabItem value="Kubernetes" label="Kubernetes">
 
-| **Name**                       | **Type**  | **Default** | **Usage**    |
+| **Annotation name**            | **Type**  | **Default** | **Usage**    |
```
wasaga, line 50:
```suggestion
| [**Annotation name**](/docs/deploy/k8s/ingress#annotations)            | **Type**  | **Default** | **Usage**    |
```

wasaga (changes requested):
What I suggest is

1. link parameter header to CRD explanation
2. link annotation header to Ingress annotation explanations

wasaga (approved):

PR #1138, opened 2024-01-05
Docs: adds allow sign-out URL step to Cognito

### LINKED ISSUE #1134: Docs: Update Cognito guide for v25
Update the Cognito IdP guide to inform users to allow the following sign-out URL: `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out`

Related to this [comment](https://github.com/pomerium/pomerium/pull/4896#discussion_r1442322165).

ZPain8464:
Fixes #1134

### DIFF
--- a/content/docs/identity-providers/cognito.mdx
+++ b/content/docs/identity-providers/cognito.mdx
@@ -75,8 +75,8 @@ If you need to make changes after creating your pool, be aware that some setting
    | **Field** | **Description** |
    | --- | --- |
    | Enabled Identity Providers | Choose **Cognito User Pool**, unless you have set up another **Identity Provider** (eg SAML) |
-   | Allowed sign-out URLs | https://authenticate_domain/.pomerium/signed_out |
-   | Callback URL(s) | https://authenticate_service_url/oauth2/callback |
+   | Allowed sign-out URLs | `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed_out` |
+   | Callback URL(s) | `https://authenticate_service_url/oauth2/callback` |
    | Allowed OAuth Flows | Authorization code grant |
    | Allowed OAuth Scopes | Email, OpenID, Profile |
 
--- a/content/docs/identity-providers/cognito.mdx
+++ b/content/docs/identity-providers/cognito.mdx
@@ -85,6 +85,14 @@ If you need to make changes after creating your pool, be aware that some setting
 
 You can choose whether to use your own **Domain Name**, or use an AWS-provided one. The AWS-provided domain names are in the format `https://${DOMAIN-PREFIX}.auth.${AWS-REGION}.amazoncognito.com`
 
+### Configure allowed sign-out URLs
+
+Add the special Pomerium sign-out URL to Cognito's **Allowed sign-out URLs** list:
+
+`https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out`
+
+If you don't add the Pomerium sign-out URL to Cognito's allowed list, you will encounter an error when signing out of your application.
+
 ## Pomerium Configuration
 
 Once you have configured AWS Cognito, configure Pomerium to connext to it:
--- a/content/docs/identity-providers/cognito.mdx
+++ b/content/docs/identity-providers/cognito.mdx
@@ -91,7 +91,7 @@ Add the special Pomerium sign-out URL to Cognito's **Allowed sign-out URLs** lis
 
 `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out`
 
-If you don't add the Pomerium sign-out URL to Cognito's allowed list, you will encounter an error when signing out of your application.
+If you don't add the Pomerium sign-out URL to Cognito's allowed list, you will encounter an error when signing out of Cognito.
 
 ## Pomerium Configuration
 
--- a/content/docs/identity-providers/cognito.mdx
+++ b/content/docs/identity-providers/cognito.mdx
@@ -89,7 +89,7 @@ You can choose whether to use your own **Domain Name**, or use an AWS-provided o
 
 Add the special Pomerium sign-out URL to Cognito's **Allowed sign-out URLs** list:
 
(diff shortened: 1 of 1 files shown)


content/docs/identity-providers/cognito.mdx
```diff
@@ -85,6 +85,14 @@ If you need to make changes after creating your pool, be aware that some setting
 
 You can choose whether to use your own **Domain Name**, or use an AWS-provided one. The AWS-provided domain names are in the format `https://${DOMAIN-PREFIX}.auth.${AWS-REGION}.amazoncognito.com`
 
+### Configure allowed sign-out URLs
+
+Add the special Pomerium sign-out URL to Cognito's **Allowed sign-out URLs** list:
+
+`https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out`
+
+If you don't add the Pomerium sign-out URL to Cognito's allowed list, you will encounter an error when signing out of your application.
```
ZPain8464, line 94:
```suggestion
If you don't add the Pomerium sign-out URL to Cognito's allowed list, you will encounter an error when signing out of Cognito.
```

content/docs/identity-providers/cognito.mdx
```diff
@@ -85,6 +85,14 @@ If you need to make changes after creating your pool, be aware that some setting
 
 You can choose whether to use your own **Domain Name**, or use an AWS-provided one. The AWS-provided domain names are in the format `https://${DOMAIN-PREFIX}.auth.${AWS-REGION}.amazoncognito.com`
 
+### Configure allowed sign-out URLs
+
+Add the special Pomerium sign-out URL to Cognito's **Allowed sign-out URLs** list:
+
+`https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out`
```
kenjenkins, line 92: I think this should really be `signed_out`, with an underscore rather than a hyphen.

ZPain8464: @kenjenkins were you able to confirm that `signed_out` is correct?

kenjenkins: It's `signed_out` in the handler endpoint code here: https://github.com/pomerium/pomerium/blob/c84a251c933dd674211d0a96605e56abcb63fde1/authenticate/handlers.go#L91.

Denis has updated it in in his pending changelog PR as well: https://github.com/pomerium/pomerium/pull/4896/commits/4dd88595f1a42fedc426605564182bc62afa60f2.

ZPain8464, line 92:
```suggestion
`https://{AUTHENTICATE_DOMAIN}/.pomerium/signed_out`
```

content/docs/identity-providers/cognito.mdx
```diff
@@ -85,6 +85,14 @@ If you need to make changes after creating your pool, be aware that some setting
 
 You can choose whether to use your own **Domain Name**, or use an AWS-provided one. The AWS-provided domain names are in the format `https://${DOMAIN-PREFIX}.auth.${AWS-REGION}.amazoncognito.com`
 
+### Configure allowed sign-out URLs
```
kenjenkins, line 88: Sorry, one other question: would it make more sense for this to be included in the table with the other client settings (in step 3)? It seems similar to the "Callback URL(s)" setting. (Note: I haven't actually seen the Cognito settings page, so I don't actually know what order these settings appear in there.)

ZPain8464: @kenjenkins I thought that would've been a good place for the URL as well, but like you I haven't been able to see the settings page. I referenced @wasaga 's screenshot in https://github.com/pomerium/pomerium-zero/issues/1385. I think this would be the cleaner option, though.

kenjenkins: I created my own AWS account and tried to go through the process of setting up a user pool in Cognito. Here's a screenshot from what I think is the relevant settings page:

Screen Shot 2024-01-08 at 1 48 25 PM

Based on that screenshot, I think it would make sense to put 'Allowed sign-out URLs' right after the 'Allowed callback URLs' row (line 78 above).

It looks like some of the other settings have been renamed ('OAuth 2.0 grant types' instead of 'Allowed OAuth Flows' and 'OpenID Connect scopes' instead of 'Allowed OAuth Scopes'). I'll file a separate issue for that.

ZPain8464:
@kenjenkins I'm aware there is some inconsistency with our example URLs in this guide. We can address that in the Cognito refresh ticket. For now, I just want to add this documentation for v0.25.

content/docs/identity-providers/cognito.mdx
```diff
@@ -75,7 +75,8 @@ If you need to make changes after creating your pool, be aware that some setting
    | **Field** | **Description** |
    | --- | --- |
    | Enabled Identity Providers | Choose **Cognito User Pool**, unless you have set up another **Identity Provider** (eg SAML) |
-   | Callback URL(s) | https://authenticate_service_url/oauth2/callback |
+   | Allowed sign-out URLs | `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed_out` |
+   | Callback URL(s) | `https://authenticate_service_url/oauth2/callback` |
```
kenjenkins, line 79: We should pick a single placeholder for the authenticate domain here, right? Let's pick just one of either `{AUTHENTICATE_DOMAIN}` or `authenticate_service_url` and use it consistently.

```suggestion
   | Allowed callback URLs | `https://{AUTHENTICATE_DOMAIN}/oauth2/callback` |
   | Allowed sign-out URLs | `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed_out` |
```

It might even make sense to emphasize that users need to replace this placeholder with the actual domain they're using, e.g. on line 73:

> In the settings for **Pomerium** app, put in the following details (replace `{AUTHENTICATE_DOMAIN}` with the actual domain you will use for the [Authenticate Service URL](https://www.pomerium.com/docs/reference/authenticate-service-url)):

ZPain8464: @kenjenkins good call, I implemented it.

kenjenkins (approved):

PR #1145, opened 2024-01-09
Docs: rewrites wildcard processing order example

ZPain8464:
This documentation was added in v0.22, so I added a backport for each  version. Please let me know if the new example makes sense, or if it's incorrect. Suggestions welcome. 

Fixes https://github.com/pomerium/internal/issues/1670
Related to https://github.com/pomerium/internal/issues/1669

### DIFF
--- a/content/docs/reference/routes/from.mdx
+++ b/content/docs/reference/routes/from.mdx
@@ -113,21 +113,38 @@ routes:
     to: tcp://example.com:22
 ```
 
-:::tip **Note**
+### Wildcard processing behavior
 
-Wildcard From Routes (`*`) may take precedence over non-wildcard routes regardless of their ordering, and more specific wildcard routes will take precedence over less specific wildcard routes.
+Pomerium processes routes in the order they are defined in the configuration file. This holds true for external wildcard routes, too.
 
-In the example below, **Route C** will take priority over routes **A** or **B**:
+For example, given the wildcard routes below, if you send a request to `bab.example.com`, Pomerium would redirect the request to `internal_1.com`.
+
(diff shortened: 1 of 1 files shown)


content/docs/reference/routes/from.mdx
```diff
@@ -113,21 +113,38 @@ routes:
     to: tcp://example.com:22
 ```
 
-:::tip **Note**
+### Wildcard processing behavior
 
-Wildcard From Routes (`*`) may take precedence over non-wildcard routes regardless of their ordering, and more specific wildcard routes will take precedence over less specific wildcard routes.
+Pomerium processes routes in the order they are defined in the configuration file. This holds true for external wildcard routes, too.
```
calebdoxsey, line 118: I think the original note was intended to handle this case:

```yaml
routes:
  - from: https://*.example.com
    to: http://1.example.com
  - from: https://www.example.com
    to: http://2.example.com
```

It would appear that route (2) should never be matched as route (1) will always be matched first. However due to the way we setup routing, route (2) may actually get matched, even though it's ordered after route (1). 

Sorry for the confusion.

To put it another way:

> Pomerium processes routes in the order they are defined in the configuration file. However routes which do not contain wildcards (`*`) may take precedence over routes which do contain wildcards.

ZPain8464: Thanks for the clarification, and that makes sense to me. I'll update this to note accordingly.

calebdoxsey (approved):

ZPain8464:
@desimone and/or @kenjenkins either of you care to review this before merging?

desimone:
I'm not close enough to this to give a meaningful review sorry

ssveta7ak (approved):

PR #1156, opened 2024-01-10
Docs: adds core upgrade guide

ZPain8464:


desimone (approved):

content/docs/deploy/core/upgrading.mdx
```diff
@@ -13,6 +13,26 @@ sidebar_position: 10
 
 # Upgrade guide
 
+## 0.25.0
+
+### Breaking
+
+#### Base64-encoded Certificates
+
+Previously, the `certificates` key supported base64-encoded certificates as a value (this option was not documented). We've removed support for base64-encoded certificates for this setting and now require that you only use the certificate file location. To avoid incompatabilities between versions, please update these values accordingly.
```
kenjenkins, line 22: I think cspell is correct here; this should be "incompatibilities".

```suggestion
Previously, the `certificates` key supported base64-encoded certificates as a value (this option was not documented). We've removed support for base64-encoded certificates for this setting and now require that you only use the certificate file location. To avoid incompatibilities between versions, please update these values accordingly.
```

cspell.json
```diff
@@ -66,6 +66,7 @@
     "hsts",
     "httpbin",
     "identityprovider",
+    "incompatabilities",
```
kenjenkins, line 69:
```suggestion
```

kenjenkins (approved):

PR #1169, opened 2024-01-10
Docs: fixes console formatting

ZPain8464:


kenjenkins (approved):
It looks like the v0.24 section goes one level deeper and uses ##### (h5) for the individual items. Do we want v0.25 to be consistent with that?

PR #1172, opened 2024-01-11
Docs: removes cookie secure example from Enterprise tab

### LINKED ISSUE #1170: Update Cookie Secure reference: removed from Enterprise
Page: https://www.pomerium.com/docs/reference/cookies#cookie-secure

The Cookie Secure option is removed from Enterprise as of v0.25.0:

Screen Shot 2024-01-10 at 5 02 34 PM

## What's the resolution?

Let's update [this section](https://github.com/pomerium/documentation/blob/db0114f8266bfd949737e7224941060bcc9be614/content/docs/reference/cookies.mdx#L196) to mention that this setting is no longer supported in Enterprise.

ZPain8464:
Fixes #1170

### DIFF
--- a/content/docs/reference/cookies.mdx
+++ b/content/docs/reference/cookies.mdx
@@ -197,7 +197,7 @@ COOKIE_SECURE=false
 </TabItem>
 <TabItem value="Enterprise" label="Enterprise">
 
-As of v0.25, we've removed support for the **Cookie Secure** setting in the Enterprise Console (in the Console, this setting was called **Javascript Security**). Now, this setting is always enabled by default.
+As of v0.25, we've removed the **Cookie Secure** setting from the Enterprise Console (in the Console, this setting was called **HTTPS only**).
 
 </TabItem>
 <TabItem value="Kubernetes" label="Kubernetes">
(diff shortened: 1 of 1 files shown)


content/docs/reference/cookies.mdx
```diff
@@ -193,13 +197,7 @@ COOKIE_SECURE=false
...
 <TabItem value="Enterprise" label="Enterprise">
 
-Configure **Cookie Secure** with the **HTTPS only** toggle button in the Console. The button has three states:
-
-- **Unset** ("-") uses the value in your configuration file
-- **Checkmark** sets `cookie_secure` to `true`
-- **Empty** sets `cookie_secure` to `false`
-
-![Configure Cookie Secure in the Console](./img/cookies/cookie-settings.png)
+As of v0.25, we've removed support for the **Cookie Secure** setting in the Enterprise Console (in the Console, this setting was called **Javascript Security**). Now, this setting is always enabled by default.
```
kenjenkins, line 200: As far as I can tell, the part "Now, this setting is always enabled by default" is not correct. I think the UI was removed but not the backend logic, so Console may continue to control this setting (using whatever value is still in the database).

```suggestion
As of v0.25, we've removed the **Cookie Secure** setting from the Enterprise Console (in the Console, this setting was called **HTTPS only**).
```

kenjenkins (approved):

PR #1183, opened 2024-01-19
Docs: azure update

ZPain8464:
This PR updates the Azure AD idp guide. The update includes:
- Changing all instances of Azure AD to Entra ID (where appropriate. You still Azure to configure the application.)
- Changing Office 365 to Microsoft 365
- Updating screenshots and deleting the `/azure/` image directory and replacing it with a new `entra` directory

One thing I need team input on:
- Should we update the path from `/identity-providers/azure` to `/identity-providers/entra`? If so, we'd need to add redirects. I'm for keeping `azure`,  just because redirects become difficult to maintain over time. This page is also indexed as `azure`, and updating the path could have unforeseen consequences with search engines.


content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
 ---
-title: Azure Active Directory
-sidebar_label: Azure AD
+title: Entra ID (Azure Active Directory)
```
kenjenkins, line 2: Based on https://www.microsoft.com/en-us/security/business/identity-access/microsoft-entra-id it sounds like the full name is "Microsoft Entra ID".

It might also be good to put "formerly" in the title as well as the description?

```suggestion
title: Microsoft Entra ID (formerly Azure Active Directory)
```

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
+sidebar_label: Entra ID
```
kenjenkins, line 3: I think we should keep "Azure AD" in the sidebar for some time.

```suggestion
sidebar_label: Microsoft Entra ID (Azure AD)
```

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
+description: Learn how to configure Microsoft Entra ID (formally Azure AD) as an identity provider that works with Pomerium Core and Enterprise.
```
kenjenkins, line 4:
```suggestion
description: Learn how to configure Microsoft Entra ID (formerly known as Azure Active Directory) as an identity provider that works with Pomerium Core and Enterprise.
```

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
+  ]
+pagination_prev: null
 pagination_next: null
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-This page will help you configure Pomerium to use your Azure Active Directory as an IdP. It assumes you have already [installed Pomerium](/docs/deploy/)
+# Microsoft Entra ID (Azure Active Directory)
```
kenjenkins, line 24: I think we can remove this explicit heading line and use the default title instead?

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
+
+Learn how to configure Pomerium to use **Microsoft Entra ID** (formally Azure Active Directory) as an identity provider.
```
kenjenkins, line 26:
```suggestion
Learn how to configure Pomerium to use **Microsoft Entra ID** (formerly Azure Active Directory) as an identity provider.
```

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
+- Create a free [Microsoft 365 administration account](https://www.microsoft.com/en-us/microsoft-365/business/office-365-administration) (or have admin privileges with an existing Microsoft 365 account)
 
-## Create a New Application
+## Access your Entra ID instance
 
-1. Login to Microsoft Azure and choose **Azure Active Directory**. You may need to click on **More Services** first:
+If you want users to sign in to an application using a Microsoft Entra ID account, either from your organization or from external directories, you must register the application through the Microsoft Azure portal.
 
-   ![Select Active Directory](./img/azure/azure-dashboard.png)
+If you have a Microsoft 365 account (formally Office 365), you can use your account's Entra ID instance instead of creating a new one. You can access Entra ID from the **Microsoft 365 admin center** or directly from the **Azure portal**.
```
kenjenkins, line 45:
```suggestion
If you have a Microsoft 365 account (formerly Office 365), you can use your account's Entra ID instance instead of creating a new one. You can access Entra ID from the **Microsoft 365 admin center** or directly from the **Azure portal**.
```

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,177 @@
...
 ```
 
 </TabItem>
 </Tabs>
 
+:::note
+
+As of this writing, Pomerium only supports **`azure`** as the valid **`idp_provider`** name to reference Microsoft Entra ID as an identity provider (not **`entra`**).
+
+This means you **do not** need to update **`azure`** in your configuration file. If you update the **`idp_provider`** key to **`entra`**, it will break your configuration.
```
kenjenkins, line 154: To me, the way this is phrased feels like we're assuming users have an existing configuration file with `azure` in it already. Given that this page is framed as a guide, I don't think this makes sense.

I'd suggest we stick to something simpler here:

```suggestion
Pomerium currently uses **`azure`** as the **`idp_provider`** name to refer to Microsoft Entra ID.
```

content/docs/identity-providers/azure.mdx
ZPain8464: So this is more of an SEO thing. The `title` front matter is treated as the title tag (what you'd see in the browser tab). If you don't include an H1 like I have, Docusaurus will use the `title` front matter value as the H1. Specifically for this guide, including both is redundant, but sometimes the H1 should differ from the title (like the code-server guide). I do this out habit to make sure future pages are treated the same.

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,175 @@
...
+## Before you begin
 
-There is no way to create an application that integrates with Microsoft Azure AD without having **your own** Microsoft Azure AD instance.
-
-:::
+To complete this guide, you must:
 
-If you have an Office 365 account, you can use the account's Azure AD instance instead of creating a new one. To find your Office 365 account's Azure AD instance:
+- Install [Pomerium](/docs/deploy/core)
+- Create a free [Microsoft 365 administration account](https://www.microsoft.com/en-us/microsoft-365/business/office-365-administration) (or have admin privileges with an existing Microsoft 365 account)
```
calebdoxsey, line 39: I think you sign up for an Azure account?

https://azure.microsoft.com/en-us/free/

ZPain8464: You can do that as well. This article explains it a bit better: https://learn.microsoft.com/en-us/microsoft-365/enterprise/azure-integration?view=o365-worldwide

calebdoxsey (approved):

content/docs/identity-providers/azure.mdx
kenjenkins: Do you want to update this heading to match the current title then?

kenjenkins (approved):

content/docs/identity-providers/azure.mdx
```diff
@@ -1,156 +1,175 @@
...
+  ]
+pagination_prev: null
 pagination_next: null
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-This page will help you configure Pomerium to use your Azure Active Directory as an IdP. It assumes you have already [installed Pomerium](/docs/deploy/)
+# Microsoft Entra ID (Azure Active Directory)
```
ZPain8464, line 24:
```suggestion
# Microsoft Entra ID (formerly Azure Active Directory)
```

PR #1185, opened 2024-01-22
Docs: deprecates cookie secure

ZPain8464:
Resolves https://github.com/pomerium/internal/issues/1462
Resolves https://github.com/pomerium/internal/issues/1682

### DIFF
--- a/content/docs/reference/cookies.mdx
+++ b/content/docs/reference/cookies.mdx
@@ -163,7 +163,7 @@ cookie:
 
 :::caution
 
-As of v0.25, we have deprecated the **Cookie Secure** setting . Pomerium will always set the `Secure` attribute on session cookies going forward. 
+As of **v0.25**, we have deprecated the **Cookie Secure** setting. Pomerium will always set the `Secure` attribute on session cookies going forward.
 
 :::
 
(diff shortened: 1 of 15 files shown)


ZPain8464:
@kenjenkins this is ready for a review

content/docs/reference/cookies.mdx
```diff
@@ -161,6 +161,12 @@ cookie:
 
 ## Cookie Secure {#cookie-secure}
 
+:::caution
+
+As of v0.25.0, we have deprecated the **Cookie Secure** setting in Pomerium Core, the Enterprise Console, and the Ingress Controller and will remove it in **v0.26.0**. It will be enabled by default going forward.
```
kenjenkins, line 166: nit: the phrase "by default" doesn't seem quite accurate if you can't change it at all; maybe something like "Pomerium will always set the `Secure` attribute on session cookies going forward."

kenjenkins, line 166: Do you think it helps to mention all of Core, Enterprise Console, and Ingress Controller?

I wonder if it would be simpler just to say: "As of v0.25, the **Cookie Secure** setting is deprecated."

content/docs/reference/cookies.mdx
```diff
@@ -180,38 +186,17 @@ This setting cannot be set to `false` if [**Cookie SameSite**](#cookie-samesite)
...
-```bash
-COOKIE_SECURE=false
-```
+As of **v0.25.0**, we've deprecated the **Cookie Secure** setting in Pomerium Core and will remove it in **v0.26.0**. Now, this setting will be enabled by default.
 
 </TabItem>
 <TabItem value="Enterprise" label="Enterprise">
 
-As of v0.25, we've removed the **Cookie Secure** setting from the Enterprise Console (in the Console, this setting was called **HTTPS only**).
+As of **v0.25.0**, we've removed the **Cookie Secure** setting from the Enterprise Console (in the Console, this setting was called **HTTPS only**). Now, this setting will be enabled by default.
```
kenjenkins, line 194: nit: adding the ".0" here may be misleading; the UI was removed in Console v0.25.0 but the corresponding backend change is pending a v0.25.1 patch release.

In general, I think it might be better to reference just the minor release (v0.25) unless we're referring specifically to a patch release.

content/docs/reference/cookies.mdx
```diff
@@ -180,38 +186,17 @@ This setting cannot be set to `false` if [**Cookie SameSite**](#cookie-samesite)
 <Tabs>
 <TabItem value="Core" label="Core">
 
-| **Config file keys** | **Environment variables** | **Type**  | **Default** |
```
kenjenkins, line 183: I don't think it makes sense to remove the setting reference here in the v0.25 docs.

content/docs/reference/cookies.mdx
ZPain8464, line 166:
```suggestion
As of v0.25, we have deprecated the **Cookie Secure** setting . Pomerium will always set the `Secure` attribute on session cookies going forward. 
```

content/docs/reference/cookies.mdx
ZPain8464, line 194:
```suggestion
As of **v0.25**, we've removed the **Cookie Secure** setting from the Enterprise Console (in the Console, this setting was called **HTTPS only**). 
```

content/docs/reference/cookies.mdx
```diff
@@ -161,6 +161,12 @@ cookie:
 
 ## Cookie Secure {#cookie-secure}
 
+:::caution
+
+As of **v0.25**, we have deprecated the **Cookie Secure** setting. Pomerium will always set the `Secure` attribute on session cookies going forward.
```
kenjenkins, line 166: Sorry, I didn't mean to imply that you should delete the "and will remove it in **v0.26**" part. The phrasing here might be a little confusing as to the behavior in v0.25.

What do you think about this?
```suggestion
As of **v0.25** the **Cookie Secure** setting is deprecated, and it will be removed in **v0.26**. Once removed, Pomerium will always set the `Secure` attribute on session cookies.
```

kenjenkins (approved):

content/docs/reference/cookies.mdx
ZPain8464: That's fine. This was a weird process for this setting, so I appreciate the help with the messaging.

PR #1198, opened 2024-01-24
Docs: add enterprise admonition

ZPain8464:
Resolves https://github.com/pomerium/internal/issues/1696

This PR presents a first draft of the custom Enterprise admonition. For now, we're using an MUI icon as a stopgap until we decide on a different (or more custom) icon.


desimone:
Can you link to a few example pages for context @ZPain8464

ZPain8464:
@desimone I replaced the admonitions in Original User Context and Branding. I added a new one to PPL underneath the Criteria section.

ZPain8464:
@kenjenkins I'm open to any icon here provided we all agree on it. The business one looks "enterprise-y" to me, I'd be fine with that.

ZPain8464:
@desimone what do you think of the business icon @kenjenkins proposed? @nhayfield if I could get your blessing on this as well that'd be great.

nhayfield (approved):
Lgtm

ssveta7ak:
Looks good for me. Also, I like the Mui business icon.

src/theme/Admonition/types.css
```diff
@@ -0,0 +1,27 @@
+.custom_admonition {
+  border: #ededed 1px solid;
+  border-left: #6F43E7 solid 4px;
+  border-radius: 10px;
+  margin: 0 0 16px 0;
+  padding: 16px;
+  background-color: #f6f3ff;
```
kenjenkins, line 7: I think we'll need a different set of colors for dark mode, as currently this has poor text contrast:

Screen Shot 2024-01-30 at 12 20 55 PM

src/theme/Admonition/Types.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import DefaultAdmonitionTypes from '@theme-original/Admonition/Types';
+import BusinessIcon from '@mui/icons-material/Business';
+
+import './types.css';
```
kenjenkins, line 5: If we're going to add a separate CSS file for this component, can it be scoped to just this component? (See [src/theme/Footer/Logo/index.js](https://github.com/pomerium/documentation/blob/e8fd25bef95706c125c255b01d9fa7baa265a49a/src/theme/Footer/Logo/index.js#L10) for an example of using a CSS module.)

ZPain8464:
@kenjenkins would you mind reviewing with the dark mode color palette?

kenjenkins (approved):
Thanks, the dark mode colors look much better to me.

I think the contrast may still be a little low... https://webaim.org/resources/contrastchecker/?fcolor=E3E3E3&bcolor=7161EF puts the text color contrast ratio at 3.51:1. 

If we darken the background to say #4d25bb, we can bring that up to above 7: https://webaim.org/resources/contrastchecker/?fcolor=E3E3E3&bcolor=4D25BB.

Also, as discussed over Slack, I've gone ahead and simplified the component a little using `[data-theme]` rules in the CSS. (I also took the liberty of renaming the CSS module to styles.module.css, as that seems to be the convention from other components with style modules.)

Issue #1205, opened 2024-01-26
"Idle Timeout" fills up settings reference table when changing sort order


kenjenkins:
Page: https://www.pomerium.com/docs/reference

## What's incorrect or missing

"Idle Timeout" can take over the entire settings reference data table:

<img width="1067" alt="Screen Shot 2024-01-26 at 2 43 43 PM" src="https://github.com/pomerium/documentation/assets/51246568/8f3fcce2-e558-4975-a444-ff3dc0bba7ce">

Steps to reproduce:
1. Navigate to https://www.pomerium.com/docs/reference.
2. Repeatedly tap the sort arrow next to the 'Name' column header.

## What's the resolution?

I suspect this might have something to do with the IDs in the reference.json file: the entries "idle-timeout" and "route-idle-timeout" both have the same ID "idle-timeout".

ZPain8464:
Thanks for catching this.

PR #1206, opened 2024-01-29
Docs: updates route-level idle timeout ID

ZPain8464:
I know we've documented internally how we document our route- and global-level conventions, but this setting predates those decisions. For now, I've updated the route-level idle timeout setting ID to `route-idle-timeout`, as that was the original name given to the route-level setting. The Console and Zero repos' help docs also label this setting "route-idle-timeout."

I've filed issues on both repos to update the ID:
- https://github.com/pomerium/pomerium-zero/issues/1564
- https://github.com/pomerium/pomerium-console/issues/3998

Fixes https://github.com/pomerium/documentation/issues/1205

### DIFF
--- a/content/docs/reference/reference.json
+++ b/content/docs/reference/reference.json
@@ -868,7 +868,7 @@
     "path": "/routes/timeouts"
   },
   "route-idle-timeout": {
-    "id": "idle-timeout",
+    "id": "route-idle-timeout",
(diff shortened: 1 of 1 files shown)


nhayfield (approved):
The ID is not used in console or zero. Only the outer key is used. No updates are needed for console and zero.

PR #1208, opened 2024-01-30
Docs: adds robots.txt notice

ZPain8464:
Fixes https://github.com/pomerium/internal/issues/1693

### DIFF
--- a/content/docs/reference/routes/public-access.mdx
+++ b/content/docs/reference/routes/public-access.mdx
@@ -22,7 +22,7 @@ Because the **Public Access** setting bypasses authentication and authorization 
 
 :::
 
-**Public Access** allows all requests to a given route and bypasses authentication and authorization. If you enable this setting, no whitelists should be provided for the route (for example, **Allowed Users**).
+**Public Access** allows all requests to a given route and bypasses authentication and authorization. If you enable this setting, no other policy should be provided for the route.
 
 ### Robots.txt behavior
 
--- a/content/docs/reference/routes/public-access.mdx
+++ b/content/docs/reference/routes/public-access.mdx
@@ -26,7 +26,7 @@ Because the **Public Access** setting bypasses authentication and authorization 
 
 ### Robots.txt behavior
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/routes/public-access.mdx
```diff
@@ -16,9 +16,24 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
-**Use with caution:** **Public Access** allows all requests for a given route, bypassing authentication and authorization. Suitable for publicly exposed web services.
+:::caution
 
-If this setting is enabled, no whitelists (for example, Allowed Users) should be provided in this route.
+Because the **Public Access** setting bypasses authentication and authorization checks, you should only enable it for publicly exposed web services.
```
kenjenkins, line 21: nit:

```suggestion
Because the **Public Access** setting bypasses authentication and authorization checks, you should enable it only for publicly exposed web services.
```

(This might be my <span>#</span>1 writing style nit ever since I became aware of it. See e.g. https://getitwriteonline.com/placement-of-only/ for more on the rationale.)

content/docs/reference/routes/public-access.mdx
```diff
@@ -16,9 +16,24 @@ import TabItem from '@theme/TabItem';
...
+
+:::
+
+**Public Access** allows all requests to a given route and bypasses authentication and authorization. If you enable this setting, no whitelists should be provided for the route (for example, **Allowed Users**).
```
kenjenkins, line 25: I think "Allowed Users" was a setting that existed before PPL. I'm not sure that it's still supported, but even if it is I don't think it's documented, so I don't think it makes sense to refer to it here.

Maybe something like this instead?
```suggestion
**Public Access** allows all requests to a given route and bypasses authentication and authorization. If you enable this setting, no other policy should be provided for the route.
```

kenjenkins, line 25: Also, this feels a little repetitive to me, compared with the previous paragraph. Is there a way to make this a little smoother?

kenjenkins (approved):
Sure, looks good to me.

PR #1211, opened 2024-01-30
Docs: updates duplicate/missing references

ZPain8464:
This PR:
- Adds fragments to all `identity-provider-settings` paths
- Removes the `client-certificate-authority` entry
- Updates `polling-max-delay` description to use `maximum`  instead of `minimum`

**Note**: One discrepancy is the `identity-provider-polling-min-max-delay` ID in `reference.json`, which isn't consistent with the anchor link:  The anchor link is `#identity-provider-minmax-delay`, so the `path` uses this fragment instead of the key/id found in the reference file.

Resolves https://github.com/pomerium/internal/issues/1701

### DIFF
--- a/content/docs/reference/reference.json
+++ b/content/docs/reference/reference.json
@@ -129,14 +129,6 @@
     "short_description": "Certificate relative file location string",
     "services": []
   },
-  "client-certificate-authority": {
-    "id": "client-certificate-authority",
-    "title": "Client Certificate Authority",
-    "path": "/certificates",
-    "services": [],
-    "short_description": "Client certificate authority relative file location string",
-    "type": "string"
-  },
   "downstream-mtls-client-certificate-authority": {
     "id": "downstream-mtls-client-certificate-authority",
     "title": "Downstream mTLS Certificate Authority",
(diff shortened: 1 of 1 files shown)


content/docs/reference/reference.json
```diff
@@ -129,14 +129,6 @@
     "short_description": "Certificate relative file location string",
     "services": []
   },
-  "client-certificate-authority": {
```
kenjenkins, line 132: I think it would be better to remove this entry in a separate PR, if this PR will be backported to the v0.25 docs site. The corresponding change in Core (https://github.com/pomerium/pomerium/pull/4918) won't be released until v0.26 at the earliest.

kenjenkins (approved):

content/docs/reference/reference.json
ZPain8464: @kenjenkins thanks for letting me know. I don't see a purpose in backporting these changes unless we want the idp fragments to be available in v25 for some reason?

kenjenkins: It's nice for the reference table entries to link directly to the corresponding setting documentation, but I agree it's not a major issue, so if you want to skip backporting this PR I think that would be fine as well.

PR #1215, opened 2024-02-01
Update Enterprise admonitions

ZPain8464:
This PR mainly updates old admonitions, but it also adds a few:
- Tooljet guide
- Jenkins guide
- Integrations page
- Authorization
- Authentication
- Device identity

I should add this is just the initial use case for the admonition. I'd like to strategically place it with more compelling text to help broadcast Enterprise features and use cases.

Resolves https://github.com/pomerium/internal/issues/1711

### DIFF
--- a/content/docs/capabilities/authentication.mdx
+++ b/content/docs/capabilities/authentication.mdx
@@ -54,6 +54,10 @@ By configuring your applications to route requests to Pomerium’s Proxy service
 
 ## External data sources (Enterprise)
 
+:::enterprise Pomerium Enterprise
+
 [Enterprise customers](https://www.pomerium.com/enterprise-sales/) can enforce context-aware access with Pomerium’s [external data sources](/docs/integrations) feature (directory sync).
+
+:::
 
 From the Enterprise Console, you can import external data from sources other than your IdP. User identity context such as users, groups, roles, language, time zones, location, and more can be included into your authorization policy so you can make granular access control decisions.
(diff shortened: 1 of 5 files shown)


content/docs/capabilities/authentication.mdx
```diff
@@ -54,6 +54,10 @@ By configuring your applications to route requests to Pomerium’s Proxy service
 
 ## External data sources (Enterprise)
 
+:::enterprise Pomerium Enterprise
```
kenjenkins, line 57: side note: now that we have a custom component for this admonition, we may want to implement a default title in it, so we wouldn't need to repeat "Pomerium Enterprise" each time we use it

content/docs/capabilities/authorization.mdx
```diff
@@ -34,10 +34,14 @@ You can apply policies in Pomerium to [Namespaces](/docs/capabilities/namespacin
 
 ### Namespaces
 
-Administrators can create a namespace, add users, groups, and routes to it, and configure a policy that applies to that specific namespace.
+:::enterprise Pomerium Enterprise
 
 Namespace support is only available for [Enterprise customers](https://www.pomerium.com/enterprise-sales/).
```
kenjenkins, line 39: optional grammar nit, while you're here:
```suggestion
Namespace support is available only for [Enterprise customers](https://www.pomerium.com/enterprise-sales/).
```

kenjenkins (approved):

content/docs/capabilities/authentication.mdx
ZPain8464: Based on their example, the title must be manually entered: https://docusaurus.io/docs/markdown-features/admonitions#custom-admonition-type-components. If it's possible to add a default title, we can do that for sure.

PR #1222, opened 2024-02-05
minor updates to the 'From Source' page

kenjenkins:
The part about "a wide array of architectures and operating systems" predates the dependency on Envoy. Currently Pomerium is limited by the OSes/architectures supported by Envoy, which I believe includes only Linux and macOS, for amd64 and arm64. (According to an announcement at https://groups.google.com/g/envoy-announce/c/2e-mSnit_BQ/m/VvweUrdaAAAJ Windows is no longer supported after Envoy 1.27.)

As of pomerium/pomerium@faa2a8652b8f001f6777d017af8f5ea021f1dc4b golangci-lint is no longer required to be installed locally (it is instead invoked via `go run`).

Finally, make sure "Pomerium" is capitalized consistently.

### DIFF
--- a/content/docs/deploy/core/from-source.mdx
+++ b/content/docs/deploy/core/from-source.mdx
@@ -20,18 +20,17 @@ import ConfigMinimal from '@site/content/examples/config/config.minimal.yaml.md'
 
 # Building Pomerium From Source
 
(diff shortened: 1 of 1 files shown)


ZPain8464 (approved):

ZPain8464:
@kenjenkins thanks for the PR :)

PR #1226, opened 2024-02-06
Updates route-level pass id headers setting

### LINKED ISSUE #1224: Pass Identity Headers: reference global option from per-route option
Page: https://www.pomerium.com/docs/reference/routes/pass-identity-headers-per-route

The reference page for the global Pass Identity Headers option (https://www.pomerium.com/docs/reference/pass-identity-headers#summary) links to the reference page for the per-route Pass Identity Headers option, but there is no link in the opposite direction.

## What's the resolution?

This page should mention the global setting, and the 'How to configure' table should not list the default as `false`.

ZPain8464:
This PR adds a link from the route-level pass identity headers setting to the global-level setting. It also removes the `default` column from the configuration table.

Fixes #1224

### DIFF
--- a/content/docs/reference/pass-identity-headers.mdx
+++ b/content/docs/reference/pass-identity-headers.mdx
@@ -18,7 +18,7 @@ import TabItem from '@theme/TabItem';
 
 When set to true, the **Pass Identity Headers** setting sends identity headers to all upstream applications.
 
-If you enable the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route) setting for a route, the route setting will take precedence over the global setting.
+To change this behavior for a specific route, use the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route).
 
 Identity headers include:
 
--- a/content/docs/reference/pass-identity-headers.mdx
+++ b/content/docs/reference/pass-identity-headers.mdx
@@ -18,7 +18,7 @@ import TabItem from '@theme/TabItem';
 
 When set to true, the **Pass Identity Headers** setting sends identity headers to all upstream applications.
 
-If a route already has the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route) setting set, the route setting will take precedence over the global setting.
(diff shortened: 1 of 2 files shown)


content/docs/reference/routes/pass-identity-headers.mdx
```diff
@@ -16,9 +16,11 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
-When set, **Pass Identity Headers** will pass identity headers to the upstream application.
+When set to true, the **Pass Identity Headers (per route)** setting will pass identity headers to the upstream application.
 
-These headers include:
+If you enable the [global-level Pass Identity Headers](/docs/reference/pass-identity-headers) setting, the route-level setting will still take precedence.
```
kenjenkins, line 21: What do you think about something like this?

```suggestion
The default value for this setting is controlled by the [global-level Pass Identity Headers](/docs/reference/pass-identity-headers) setting.
```

content/docs/reference/pass-identity-headers.mdx
```diff
@@ -18,7 +18,7 @@ import TabItem from '@theme/TabItem';
 
 When set to true, the **Pass Identity Headers** setting sends identity headers to all upstream applications.
 
-If a route already has the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route) setting set, the route setting will take precedence over the global setting.
+If you enable the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route) setting for a route, the route setting will take precedence over the global setting.
```
kenjenkins, line 21: nit: the word "enable" makes me think this is talking about setting this to `true`, but you might just as well have the global setting `true` and individual routes set to `false`. What do you think about "configure" instead of "enable" here?

```suggestion
If you configure the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route) setting for a route, the route setting will take precedence over this global setting.
```

Or to phrase this another way entirely, maybe something like this?

```suggestion
To change this behavior for a specific route, use the [route-level Pass Identity Headers](/docs/reference/routes/pass-identity-headers-per-route).
```

kenjenkins (approved):

PR #1228, opened 2024-02-06
document new cert matcher SAN options

kenjenkins:
Add the new Subject Alternative Name match options (`san_dns`, `san_email`, and `san_uri`) to the table of Certificate Matcher conditions on the PPL reference page. Add an email SAN usage example.

Related issue:
 - https://github.com/pomerium/pomerium/issues/4615

### DIFF
--- a/content/docs/capabilities/ppl.mdx
+++ b/content/docs/capabilities/ppl.mdx
@@ -168,6 +168,9 @@ A certificate matcher can be used to allow or deny specific TLS certificates, ba
 | Key Name | Value Type | Description |
 | --- | --- | --- |
 | `fingerprint` | string or array of strings | The certificate's SHA-256 fingerprint must match one of the provided values. |
+| `san_dns` | [String Matcher] | The certificate must contain a Subject Alternative Name with a DNS name satisfying the provided condition. |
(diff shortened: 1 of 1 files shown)


content/docs/capabilities/ppl.mdx
```diff
@@ -168,6 +168,9 @@ A certificate matcher can be used to allow or deny specific TLS certificates, ba
 | Key Name | Value Type | Description |
```
ZPain8464, line 168: In the paragraph above this ("A certificate matcher can be used to..."), you mention that a cert matcher can be used to allow or deny certs based on the certificate fingerprint or SPKI hash, but you do not mention the new SAN options in the table below. Was that intentionally omitted or should that be included in the criteria?

ZPain8464 (approved):

content/docs/capabilities/ppl.mdx
kenjenkins: Good catch! I think the table is descriptive enough on its own, so maybe we can simplify the introductory paragraph like this?

ZPain8464: Yes, that update makes sense to me.

ZPain8464 (approved):

PR #1230, opened 2024-02-06
Removes client ca reference page settings

ZPain8464:
This PR removes the Client Certificate Authority setting from `/docs/reference/certificates`. It also removes the entry in `reference.json` and adds a `caution` admonition with a hard-coded link to the last supported version. 

Resolves https://github.com/pomerium/internal/issues/1704

### DIFF
--- a/content/docs/reference/certificates.mdx
+++ b/content/docs/reference/certificates.mdx
@@ -150,11 +150,3 @@ Kubernetes does not support `certificate_authority`
 
 </TabItem>
 </Tabs>
-
-:::caution
-
-We've removed support for the `client_ca` and `client_ca_file` settings. To configure client certificates for downstream connections, see the **Downstream mTLS Settings** group (see the [**Certificate Authority**](/docs/reference/downstream-mtls-settings#ca) setting there).
-
-If you need the older `client_ca` and `client_ca_file` settings, please see the [**v0.25.0 Client Certificate Authority**](https://0-25-0.docs.pomerium.com/docs/reference/certificates#client-certificate-authority) documentation, or an earlier Pomerium version.
-
-:::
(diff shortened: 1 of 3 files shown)


content/docs/reference/certificates.mdx
```diff
@@ -152,52 +151,10 @@ Kubernetes does not support `certificate_authority`
 </TabItem>
 </Tabs>
 
-## Client Certificate Authority {#client-certificate-authority}
-
 :::caution
 
-This setting is deprecated; it has been moved to a new **Downstream mTLS Settings** group (see the [**Certificate Authority**](/docs/reference/downstream-mtls-settings#ca) setting there).
+We've removed support for the `client_ca` and `client_ca_file` settings. To configure client certificates for downstream connections, see the **Downstream mTLS Settings** group (see the [**Certificate Authority**](/docs/reference/downstream-mtls-settings#ca) setting there).
```
kenjenkins, line 156: I think it would be fine to remove this note entirely. I'm not sure it adds much.

If we do keep it, I would re-phrase this: I think it would be more accurate to say this setting has "moved".

content/docs/reference/certificates.mdx
```diff
@@ -152,52 +151,10 @@ Kubernetes does not support `certificate_authority`
...
 
-This setting will be treated as an alias for the new setting, but will be removed in a future Pomerium release.
+If you need the older `client_ca` and `client_ca_file` settings, please see the [**v0.25.0 Client Certificate Authority**](https://0-25-0.docs.pomerium.com/docs/reference/certificates#client-certificate-authority) documentation, or an earlier Pomerium version.
```
kenjenkins, line 158: Again, the setting has moved, but otherwise should behave just the same as in previous versions. I'm not sure it's helpful to include this link, especially if the `#client-certificate-authority` anchor no longer links here.

kenjenkins (commented):
Let's also update the previous redirect entry in the _redirects file:
https://github.com/pomerium/documentation/blob/6a0723524dd56ac8c3c8f6ced067476a224e4fc5/static/_redirects#L379

This can point to `/docs/reference/downstream-mtls-settings#ca` instead.

kenjenkins (approved):

PR #1233, opened 2024-02-06
update 'Pomerium Core (Server)'

kenjenkins:
I mainly wanted to add a note for https://github.com/pomerium/internal/issues/1707, but then I noticed there was a lot more on this page that could be updated. Let me know if you have any thoughts about any of these proposed changes.

---

Add a note about minimum glibc version (Linux) and minimum macOS version for the official binaries. Also replace "darwin" with the more common term "macOS".

Remove the link to the service mode setting. (We don't encouraging split service mode and I don't think it's helpful to mention this here.) Capitalize "Pomerium Core" consistently.

Add an item about the new `:vX.Y` docker image tags. Move the "minimal" link down to the section about debug images. Update the part about rootless container permissions/listening port so that it includes a link to the corresponding settings page. Be consistent about including `:` in the tag names. Rephrase list items to avoid putting a `:` character both in a tag name and right after it.

Remove outdated information about Helm. Replace the description front matter, as it looks to have been copied from some other page.

### DIFF
--- a/content/docs/deploy/core.mdx
+++ b/content/docs/deploy/core.mdx
@@ -94,12 +94,14 @@ Pomerium also provides [Docker container images](https://www.docker.com/resource
 
 - `:vX.Y` corresponds to the latest patch release for a specific minor version (starting with v0.25).
 
+  ```shell-session
+  $ docker pull pomerium/pomerium:v0.25
+  ```
+
 - `:latest` corresponds to the [most recent tagged release](https://github.com/pomerium/pomerium/releases/latest).
 
   ```shell-session
-  $ docker pull pomerium/pomerium:latest && docker run pomerium/pomerium:latest --version
-  pomerium: 0.25.0-1704902203+e6ed4d53
-  envoy: 1.28.0+eb930e32ab5555643e09d11d490e392d0a790c5a80eb0b0ebacb1046bdbb114d0
+  $ docker pull pomerium/pomerium:latest
   ```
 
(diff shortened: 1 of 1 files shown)


content/docs/deploy/core.mdx
```diff
@@ -79,57 +82,52 @@ echo "deb https://dl.cloudsmith.io/public/pomerium/pomerium/deb/debian buster ma
...
   ```shell-session
-  $ docker run pomerium/pomerium:v0.1.0 --version
-  v0.1.0+53bfa4e
+  $ docker run pomerium/pomerium:v0.25.0 --version
+  pomerium: 0.25.0-1704902203+e6ed4d53
+  envoy: 1.28.0+eb930e32ab5555643e09d11d490e392d0a790c5a80eb0b0ebacb1046bdbb114d
   ```
 
-- `:latest`: which will pull the [most recent tagged release](https://github.com/pomerium/pomerium/releases).
+- `:vX.Y` corresponds to the latest patch release for a specific minor version (starting with v0.25).
```
ZPain8464, line 95: I realize we don't have a v0.25 patch release for Core, but would it make sense to add an example? Like:
```shell-session
$ docker pull pomerium/pomerium:v25.X && docker run pomerium/pomerium:v25.X
```

ZPain8464 (approved):

content/docs/deploy/core.mdx
kenjenkins: Sure, but the version output looks a little repetitive if they're all using v0.25.0. How about this, where we give the example `docker pull` command but skip the version output?

PR #1245, opened 2024-02-09
Adds directory sync page

ZPain8464:
This PR adds a directory sync capabilities page. 

Resolves https://github.com/pomerium/internal/issues/1668


content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+Once you save your settings, it may take awhile for the sync to complete. Go to [**Monitor directory sync**](#monitor-directory-sync) for more information.
+
+### Monitor directory sync
+
+The Enterprise Console polls the identity provider data source based on the durations defined in the **Polling Min Delay** and **Polling Max Delay** fields.
+
+See [**Identity Provider Min/Max Delay**](/docs/reference/identity-provider-settings#identity-provider-polling-minmax-delay) for more information on how to monitor directory sync.
+
+### IDP Options
```
kenjenkins, line 36: I think most times we abbreviate "identity provider" as "IdP" (with a lowercase 'd'). I'd recommend we stick to that capitalization throughout, for consistency.

content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+:::enterprise
+
+**Directory Sync** is a Pomerium Enterprise feature. [Contact us](https://www.pomerium.com/enterprise-sales/) to upgrade today.
+
+:::
+
+## Directory sync in the Enterprise Console
+
+To start a directory sync in the Enterprise Console:
```
kenjenkins, line 20: To me the phrase "To start a directory sync" suggests this is a one-time thing, rather than a recurring process. Maybe instead: "To enable directory sync"?

content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+1. Go to the **Identity Providers** tab
+1. Select your **Identity Provider**
+1. Next to [**IDP Options**](#idp-options), fill out the required fields
```
kenjenkins, line 24: I think maybe we should add a "see below" link or something like that here. I remember being very confused by these options when I was trying to enable directory sync for testing. I think it might be helpful to have some additional framing here to explain that this varies from provider to provider.

content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
...
+1. In the [**Polling Min Delay**](#monitor-directory-sync) and [**Polling Max Delay**](#monitor-directory-sync) fields, enter the appropriate durations
```
kenjenkins, line 25: How would I choose appropriate durations? If there are reasonable defaults for these settings, maybe we could leave this step out of the main instructions?

content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
+---
+id: directory-sync
+title: Directory Sync
+description: Directory Sync in Pomerium Enterprise allows you to import organizational directory data and external data sources you can use in authorization policies.
+keywords: [directory sync, idp, identity provider, groups, directory]
+---
+
+# Directory Sync
+
+**Directory Sync** is the process of synchronizing external directory data from your identity provider into the Enterprise Console. This document discusses how directory sync works in Pomerium and its use cases.
```
kenjenkins, line 10: Is it worth mentioning up front that some of the supported IdP's do not support directory sync?

content/docs/capabilities/directory-sync.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+The Enterprise Console polls the identity provider data source based on the durations defined in the **Polling Min Delay** and **Polling Max Delay** fields.
+
+See [**Identity Provider Min/Max Delay**](/docs/reference/identity-provider-settings#identity-provider-polling-minmax-delay) for more information on how to monitor directory sync.
+
+### IDP Options
+
+The requirements and instructions for directory sync vary depending on the identity provider. You can view the **IDP Options** for an identity provider in the Enterprise Console, or refer to the relevant identity provider guide for vendor-specific steps:
+
+- [Auth0](/docs/identity-providers/auth0)
```
kenjenkins, line 40: I think it would be even better if we could link directly to the directory sync instructions on each of these pages. I realize this is complicated by the current page structure, where the instructions are often inside a tab group.

What do you think about using the [`queryString`](https://docusaurus.io/docs/next/markdown-features/tabs#query-string) prop on the tab groups, so we can link to a specific tab? e.g.
 - add `queryString="groups"` to the relevant `Tabs` element on the Auth0 page
 - change the tab `value`s to something more URL-friendly (say `oss` and `enterprise`)
 - then change this link to `/docs/identity-providers/auth0?groups=enterprise#groups`

content/docs/capabilities/directory-sync.mdx
ZPain8464: You're correct. I opted for "IDP" because that's how this field is labeled in the Enterprise Console, though. That's also why I bolded "IDP Options" and capitalized "Options" in each instance this UI element is mentioned in the text.

content/docs/capabilities/directory-sync.mdx
ZPain8464: You make a good point, but we did have a ticket awhile back where users were unaware their sync wasn't completing. It depends on the size of your directory, so the defaults don't always work. You have to monitor the sync to ensure it completes, which may require adjusting min/max polling settings to get the sync right.

content/docs/capabilities/directory-sync.mdx
ZPain8464: Definitely an enhancement. Should we do that in a separate PR, as it requires updating several IdP guides?

kenjenkins: Sure, a separate PR would be fine.

kenjenkins (approved):
Thanks, I think this is good place to start.

PR #1262, opened 2024-02-22
Adds ip range lookup support

ZPain8464:
This PR adds a section to our Integrations page explaining the CIDR indexing support for external data sources.

Resolves https://github.com/pomerium/internal/issues/1737

### DIFF
--- a/content/docs/integrations.mdx
+++ b/content/docs/integrations.mdx
@@ -121,7 +121,7 @@ CIDR indexing follows this format:
 
 ```json
 {
-  "$index": {"cidr":"1.2.3.4/5"}
+  "$index": {"cidr": "1.2.3.4/5"}
 }
 ```
 
--- a/content/docs/integrations.mdx
+++ b/content/docs/integrations.mdx
@@ -93,7 +93,7 @@ Unless defined by the directory structure of a supplied archive file, the Record
 
 #### IP range lookup support
 
-Pomerium supports Classless Inter-Domain Routing (CIDR) indexing when filtering a list of records with the `request.ip` foreign key. This allows you to match against IP address ranges instead of manually importing records for individual IP addresses.
+For the `request.ip` foreign key, Pomerium also supports matching against a range of IP addresses (expressed in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation)). This can reduce the number of data records you need.
 
 To match against an IP address range, add a special `$index` key to your external data source records. For example:
 
--- a/content/docs/integrations.mdx
+++ b/content/docs/integrations.mdx
@@ -115,7 +115,7 @@ For data sources using mTLS, you can select a [client certificate](/docs/concept
 
 ## IP range lookup support
 
-Pomerium supports Classless Inter-Domain Routing (CIDR) indexing to improve database performance. This is useful if you're importing a list of records containing IP ranges.
+Pomerium supports Classless Inter-Domain Routing (CIDR) indexing to improve database querying performance. This is useful if you're importing a list of records containing IP ranges.
 
 CIDR indexing follows this format:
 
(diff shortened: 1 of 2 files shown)


content/docs/integrations.mdx
```diff
@@ -109,6 +109,20 @@ If a job may not complete within the `min delay` period, it would be interrupted
...
 :::
 
-### **Client TLS Key**
+### Client TLS Key
 
 For data sources using mTLS, you can select a [client certificate](/docs/concepts/certificates) (added under **Manage** → **Certificates**) to provide to the data source.
+
+## IP range lookup support
+
+Pomerium supports Classless Inter-Domain Routing (CIDR) indexing to improve database querying performance. This is useful if you're importing a list of records containing IP ranges.
```
kenjenkins, line 118: The thing I was surprised about was the ability to match on IP address ranges at all (setting aside any questions to do with performance).

I'd like to see some explanation along these lines:

"When using `request.ip` as a key, you can also match against IP address ranges (expressed in CIDR notation). To do so, add a special `$index` key to your external data source records, like so:"

It also might make sense to move this up, next to the "Foreign Key" heading?

content/docs/integrations.mdx
```diff
@@ -33,7 +33,7 @@ See the pages in this section for more information on our example data sources,
 
 Any external data source integration requires the following settings:
```
kenjenkins, line 34: Side note: I don't think this is accurate. I think the last four settings are optional ("Headers", "Allow Insecure TLS", "Polling Min/Max Delay", "Client TLS Key").

ZPain8464: Yes, I was actually looking for "Allow Insecure TLS" in the Console UI and couldn't find it at all. I'll submit an issue to update this page, but it's not a high priority in my opinion at the moment.

content/docs/integrations.mdx
```diff
@@ -78,28 +78,42 @@ The Pomerium Databroker would be updated with types `example.com/geoip`, `device
...
-- `device.id`.
+- `user.id` (Also the default if no value is provided)
+- `user.email`
+- `request.ip`
+- `device.id`
 
-### **Headers**
+#### IP range lookup support
+
+Pomerium supports Classless Inter-Domain Routing (CIDR) indexing when filtering a list of records with the `request.ip` foreign key. This allows you to match against IP address ranges instead of manually importing records for individual IP addresses.
```
kenjenkins, line 96: "Pomerium supports Classless Inter-Domain Routing (CIDR) indexing" sounds off to me. "Classless Inter-Domain Routing" has to do with network architecture. "CIDR notation" is the operative term here.

I understand the desire to spell out any potentially-unfamiliar acronyms, but here I think it would tend to increase confusion rather than decrease it.

Would a link to wikipedia help to address this concern in a different way?

```suggestion
For the `request.ip` foreign key, Pomerium also supports matching against a range of IP addresses (expressed in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation)). This can reduce the number of data records you need.
```

content/docs/integrations.mdx
```diff
@@ -78,28 +78,42 @@ The Pomerium Databroker would be updated with types `example.com/geoip`, `device
...
+
+To filter with CIDR indexing, add a special `$index` key to your external data source records. For example:
```
kenjenkins, line 98:
```suggestion
To match against an IP address range, add a special `$index` key to your external data source records. For example:
```

content/docs/integrations.mdx
```diff
@@ -78,28 +78,42 @@ The Pomerium Databroker would be updated with types `example.com/geoip`, `device
...
+
+```json
+{
+  "$index": {"cidr": "1.2.3.4/5"}
```
kenjenkins, line 102: nit: this is not a sensible range; can we use something like this instead?

```suggestion
  "$index": {"cidr": "10.1.2.3/24"}
```

kenjenkins (approved):

PR #1276, opened 2024-02-29
Updates Docker pull URLs

ZPain8464:
This PR updates all instances of the Docker pull URL in our docs.  I may have missed  a few, so I'd appreciate a quick review from the team.

I backported to v0.25, but we could also backport to previous versions. Thoughts or opinions here?

### DIFF
--- a/content/docs/deploy/core.mdx
+++ b/content/docs/deploy/core.mdx
@@ -95,19 +95,19 @@ Pomerium also provides [Docker container images](https://www.docker.com/resource
 - `:vX.Y` corresponds to the latest patch release for a specific minor version (starting with v0.25).
 
   ```shell-session
-  $ docker pull pomerium/pomerium:v0.25
+  $ docker pull cr.pomerium.com/pomerium/pomerium:v0.25
   ```
 
 - `:latest` corresponds to the [most recent tagged release](https://github.com/pomerium/pomerium/releases/latest).
 
   ```shell-session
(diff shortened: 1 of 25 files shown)


content/docs/deploy/core.mdx
kenjenkins, line 1: Should this still mention Docker Hub on line 85?

content/examples/tooljet/console-compose.yaml.md
```diff
@@ -4,7 +4,7 @@ networks:
   main: {}
 services:
   pomerium:
-    image: pomerium/pomerium:v0.21.1
+    image: cr.pomerium.com/pomerium/pomerium:v0.21.1
```
kenjenkins, line 7: I noticed you updated a `v0.23.0` tag to `latest` above. Should we do the same here, or is there some reason this needs to use an older version?

kenjenkins (approved):

content/examples/tooljet/console-compose.yaml.md
ZPain8464: The guide was written to be compatible with v0.21, which requires a [service-level network attribute](https://docs.docker.com/compose/compose-file/05-services/#networks) in the Docker Compose file. We had to update the [v0.21 Quickstart](https://0-21-0.docs.pomerium.com/docs/quickstart) to do this as well. We haven't updated this guide in some time, so I thought it best to keep the current Pomerium version.

PR #1298, opened 2024-03-07
Moves databroker service def to data-storage

ZPain8464:
This PR removes the **Databroker Service** entry from the databroker reference page, modifies the text, and moves the text to the internals persistence page. 

Resolves https://github.com/pomerium/internal/issues/1748.


kenjenkins (approved):
Thanks!

Issue #1288, opened 2024-03-04
docs: links to section won't open the section if it is inside a tab that is not selected by default


wasaga:
Page: https://www.pomerium.com/docs/identity-providers/google#set-directory-permissions-for-workspaces

<img width="999" alt="image" src="https://github.com/pomerium/documentation/assets/673131/2d6d38df-16fe-4377-b624-f19e1352527c">

## What's incorrect or missing

when you open this link, you are not positioned to this chapter. instead you'll be at the top of the page 

<img width="1029" alt="image" src="https://github.com/pomerium/documentation/assets/673131/86c2134f-9208-4ca3-88de-42d4c2f632fa">

ZPain8464:
based on [Docusaurus's docs](https://www.checkpoint.com/cyber-hub/network-security/what-is-zero-trust-application-access-ztaa/), this is not possible. The only options are:
- display default tabs
- syncing tab choices
- customizing tabs
- adding a query string to the url
Adding a query string in the URL doesn't function like an anchor link -- it just pre-selects the the tab(s), but won't scroll to the tab. 

The Help Docs links do work as an anchor link, and take the user directly to the setting header, but I believe it's because the headers aren't nested in a tab.

kenjenkins:
Ken: what happens if we combine the [query string option](https://docusaurus.io/docs/markdown-features/tabs#query-string) with a URL fragment? Can we link directly to a section that way?

PR #1302, opened 2024-03-08
adds query strings to IdP guides

ZPain8464:
This PR adds query strings to Tabs. Now, when you send a link that includes an anchor link within a Tab, the page will scroll to the correct location. 

Note: I did add `queryString` to the Config and Envar Tabs, too. If you click one of these tabs, and then click another Tab in a separate  block, with a query string, it will conjoin them. For example: 

```markdown
<Tabs queryString="foo-a">
<TabItem value="bar-a"></TabItem>
<TabItem value="baz-a"></TabItem>
</Tabs>
<Tabs queryString="foo-b">
<TabItem value="bar-b"></TabItem>
<TabItem value="baz-b"></TabItem>
</Tabs>
```
Selecting `baz-a` and then `bar-b` would result in a string like:

`?foo-a=baz-a&foo-b=bar-b`. A real life example:

`/docs/identity-providers/github?configuration-settings=environment-variables&get-groups=directory-sync`

It's entirely possible that user would unwittingly select both Tabs, which would break the functionality. I want the team's input, but maybe we should only add query strings to one tab block in a page going forward? Happy to document this internally. 

Resolves https://github.com/pomerium/documentation/issues/1288

### DIFF
--- a/content/docs/identity-providers/google.mdx
+++ b/content/docs/identity-providers/google.mdx
@@ -67,7 +67,7 @@ Click **Create** once complete.
 
 Edit `config.yaml` or set your [environment variables] to connect Pomerium to Google:
 
-<Tabs queryString="configuration-settings">
+<Tabs>
 <TabItem value="config-file-keys" label="config.yaml">
 
 ```yaml title=/etc/pomerium/config.yaml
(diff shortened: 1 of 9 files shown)


ZPain8464:
@kenjenkins I've updated this to remove `queryString` from the Google guide. I'd like to test the other guides with the strings added just to see if it's helpful at all.

content/docs/identity-providers/google.mdx
```diff
@@ -90,38 +90,40 @@ IDP_CLIENT_SECRET="xxxxxx"
 
 ## Getting Groups
 
-<Tabs queryString="get-groups">
-<TabItem value="custom-claim" label="Custom Claim (Open Source)">
+### Custom claims
 
-Unfortunately, Google does not yet support getting groups data using a custom claim. Groups must be loaded by using a plugin to fetch directory information (see Enterprise's Directory Sync).
+Unfortunately, Google doesn't support getting groups data with a custom claim. To get and use groups data from your Google directory, you must conduct a directory sync in the Enterprise Console.
```
kenjenkins, line 95: I think I've mentioned this before, but "conduct a directory sync" makes this sound like a one-time event. I'd suggest we phrase this more like "enable directory sync" to help reflect that directory sync is a recurring process.

kenjenkins (approved):

content/docs/identity-providers/google.mdx
ZPain8464: I think this stems from a misunderstanding on my part of how directory sync works. I thought it was a one-time sync, I'll update. Thanks for catching that.

PR #1326, opened 2024-03-25
Adds zero quickstart draft

ZPain8464:
This PR adds the Pomerium Zero official quickstart guide, and moves the Core Quickstart to the Core directory.

related to https://github.com/pomerium/internal/issues/1739


content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
 
-<DockerCompose />
+```bash
+docker compose up -d
+```
 
-## Run Docker Compose
+Pomerium Zero will attempt to detect your **public IP** address to connect you to your cluster.
+
+But, it won't be able to because you're running Pomerium Zero in an isolated container with its own private network and IP address.
```
cmo-pomerium, line 79: I think right after this we need to showcase what someone is seeing so it's spelled out for them

"You'll see an error like the following (see attached image), and that's completely normal!"

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
+- **SSH**
+- **Verify**
+
+### Connect to SSH
+
+In a terminal, enter the following command and replace `YOUR_CLUSTER_DOMAIN_NAME` with your own:
 
 ```bash
-docker compose up
+pomerium-cli tcp ssh.<YOUR_CLUSTER_DOMAIN_NAME>.app:22
```
cmo-pomerium, line 105: Mild nit but it doesn't have the same highlight

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
+</TabItem>
+
+</Tabs>
 
 :::
 
-## Next Steps
+### Connect to Verify
+
+In the Zero Console, select the **From** route for the Verify app.
```
cmo-pomerium, line 186:
```suggestion
Now, let's test a connection to see if it works. In the Zero Console, select the **From** route for the Verify app.
```

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
+For example:
 
-If you want to try connecting Pomerium with other services, see some of our [Guides](/docs/guides).
+![Selecting the external URL for the Verify route in the Zero Console](./img/quickstart/pz-quickstart-verify-route.png)
 
-**Did you finish this quickstart guide?** We'd love to hear what you think. Get in touch with us on our [Discuss forum](https://discuss.pomerium.com/), message us on [Twitter](https://twitter.com/pomerium_io), [LinkedIn](https://www.linkedin.com/company/pomerium-inc), or check out our [Community](https://www.pomerium.com/docs/community) page.
+:::note
 
-:::caution
+Because the Verify service uses a publically accessible URL, Pomerium Zero won't be able to connect to the `jwks.json` endpoint to fetch the signing key used to verify the user's JWT.
```
cmo-pomerium, line 194:
```suggestion
You will see the following error because the Verify service uses a publicly accessible URL. In the current setup, Pomerium Zero won't be able to connect to the `jwks.json` endpoint to fetch the signing key used to verify the user's JWT.
```

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
 :::
 
-[configuration file]: /docs/reference
-[container images]: https://hub.docker.com/r/pomerium/pomerium
-[docker]: https://docs.docker.com/install/
-[docker compose]: https://docs.docker.com/compose/install/
+
+## Build your own policy
+
+In the **Policy** tab, create a new policy that instructs Pomerim to grant access only if the user's email address ends the specified value.
```
cmo-pomerium, line 203:
```suggestion
In the **Policy** tab, create a new policy that instructs Pomerim to grant access only if the user's email address ends with the specified value.
```

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,198 @@ keywords:
...
+
+![Attach the new policy to the Verify route in the Zero Console](./img/quickstart/pz-attach-policy.png)
+
+Now, if you authenticate using an email address with a domain other than the one specified in the policy, Pomerium Zero will deny the user access.
+
+## Next Steps
+
+Now that you've completed the Pomerium Zero Quickstart, build your first route:
+
+- **Pomerium Zero Fundamentals: Build Routes**
```
cmo-pomerium, line 219: This doesn't currently link to anything. Are we waiting on the fundamentals to be built too?

cmo-pomerium (commented):
Finished my review. @ZPain8464 let me know if I can clarify anything.

content/docs/quickstart.mdx
ZPain8464: Yes, we are.

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
+The Zero Quickstart shows you how to install and run Pomerium Zero in a Docker container.
+
+## Before you start
+
+- Install [Docker](https://docs.docker.com/install/) and [Docker Compose](https://docs.docker.com/compose/install/)
+- Create a (free) [**Pomerium Zero account**](https://console.pomerium.app/create-account)
+
+## Get Pomerium Zero configuration
+
+In the Zero onboarding screen, select the **Docker** tab and copy the Docker Compose configuration. (If you selected **Finish** before copying the Docker configuration, we've provided a copy below.)
```
kenjenkins, line 39: It's probably worth mentioning something here about how you get to the onboarding screen, maybe something like "after creating an account, you should see an onboarding page"?

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
 
-Run Pomerium Core with Docker containers in **under 5 minutes**.
+- **Line 8**: Replace `CLUSTER_TOKEN` with the cluster token generated for you in the onboarding config
+- **Line 15**: Replace `CLUSTER_STARTER_SUBDOMAIN` with your generated starter subdomain
 
-The Core quickstart uses Pomerium's [**Hosted Authenticate Service**](/docs/capabilities/hosted-authenticate-service), but you can also configure a [**self-hosted authenticate service**](/docs/capabilities/self-hosted-authenticate-service) to integrate with Pomerium.
+:::info Where's my **Starter Domain?**
 
-## Prerequisites
+You can find your **Starter Domain** at the top of the Zero Console in the navigation bar.
```
kenjenkins, line 80: I think we don't show the starter domain while on the onboarding page.

kenjenkins, line 80: It might be nice to give some explanation here about what the `verify` app is and why it's included here? (Do we have a docs page about the verify app? Maybe we should add one if not?)

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
-Add the configuration below to `config.yaml`:
+Next, deploy Pomerium Zero with the following command:
 
-<ConfigDocker />
+```bash
+docker compose up -d
+```
 
-## Configure Docker
+This command will deploy a local cluster running Pomerium in a Docker container. If your cluster deploys successfully, the Pomerium Zero cloud will attempt to secure a connection to your cluster.
```
kenjenkins, line 93: I think we could give some better explanation here. In particular "Pomerium Zero cloud will attempt to secure a connection to your cluster" doesn't make sense to me. Maybe we could say something about how this starts Pomerium in a mode where it connects to the cloud service?

We might also want to hold off on talking about a "cluster" just yet... this might be confusing to some folks if they assume "cluster" implies multiple replicas.

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
+
+:::
+
+## Connect to upstream services
+
+In the Zero Console, we've provisioned two starter routes, **SSH** and **Verify**, that come with their own DNS records and TLS certificates:
+
+![Viewing the two starter routes](./img/quickstart/pz-quickstart-routes.png)
+
+### Connect to SSH
```
kenjenkins, line 113: I'd recommend starting with the simpler example service first (i.e. swapping this section and the "Connect to Verify" section).

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
 ```
 
-Access the **verify route** you built in your policy: `https://verify.localhost.pomerium.io`
+:::note
 
-If you get a self-signed certificate warning, see [Handle Self-Signed Certificate Warning](/docs/internals/troubleshooting#handle-self-signed-certificate-warning) to bypass it.
+If you haven't, install [**Pomerium CLI**](/docs/deploy/clients/pomerium-cli) with any of the following methods:
 
-You should be redirected to the **verify** service. You'll see a page like this:
+<Tabs>
```
kenjenkins, line 132: I wonder if it makes more sense to link out to a different page for the pomerium-cli installation instructions... it seems a hassle to keep duplicate copies of these instructions in sync.

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
+</TabItem>
+
+</Tabs>
 
 :::
 
-## Next Steps
+### Connect to Verify
+
+In the Zero Console, select the **From** URL for the **Verify** app.
```
kenjenkins, line 199: Either above or here, I think this would definitely benefit from some explanation — what is the verify service? what does it demonstrate?

content/docs/quickstart.mdx
```diff
@@ -22,74 +22,216 @@ keywords:
...
+
+You should land on the Verify page:
+
+![The homepage of the Verify app after the user is redirected](./img/quickstart/pz-identity-verified.png)
 
-If you want to [try Enterprise](https://www.pomerium.com/enterprise-sales/), check out the [Enterprise with Docker quickstart](/docs/deploy/enterprise/quickstart).
+## Review authorization policy
```
kenjenkins, line 205: Before talking about policies and certificates, would it make sense to show how to create your own route? I think it might be good to explain how the starter domain is set up so you can add any sub-domain as a route, and the steps involved to set up a new route, i.e.
- pick a sub-domain to use with the From URL
- set up some upstream service to use with the To URL
- attach the demo policy to grant access to the route

content/docs/quickstart.mdx
ZPain8464: Good point, I'll add that in.

content/docs/quickstart.mdx
ZPain8464: We do, since Caleb implemented the change to make the hosted Verify service reachable for deployments running in Docker. Here's the current config in our onboarding flow:
```yaml
services:
  pomerium:
    image: pomerium/pomerium:v0.26.0
    ports:
      - 443:443
    restart: always
    environment:
      POMERIUM_ZERO_TOKEN: AMf-vBy2rzctrTCXBgcAIpUfLKWSme0Bs_HtTcQYb5nbFe8idwaiIt4MjemObphF_L-Pdz_pqCggHHwszIziQU4bC2nGJklhgMTfGGEtpntqfmXRBT6D_m09maHqSJy_75DrLFL2SQOgw3pjuWELm4WhtjD1Rs8aCTl7acjPJe0CKJk5zfSMZcAeFajMhKFpEk-yq_c9G9MN
      XDG_CACHE_HOME: /var/cache
    volumes:
      - pomerium-cache:/var/cache
    networks:
      main:
        aliases:
        - verify.pro-cricket-4498.pomerium.app
  verify:
    image: cr.pomerium.com/pomerium/verify:latest
    networks:
      main:
        aliases:
        - verify

networks:
  main: {}

volumes:
  pomerium-cache:
```

ZPain8464: Good point. We do not have anything I can think of that discusses the Verify app. For now, it would be easiest to link to the GH repo and add an explanation about the service and it's intended use.

content/docs/quickstart.mdx
ZPain8464: If we want to start being more consistent with terminology, this might be a good time to mention "Zero-managed mode", and that that really means Pomerium connects to the Pomerium Zero cloud?

I see your point with the mention of a cluster. I'm in favor of removing it, we just need to agree on how we want to replace it.

content/docs/quickstart.mdx
ZPain8464: Agree.

content/docs/quickstart.mdx
ZPain8464: We can [import Markdown](https://docusaurus.io/docs/markdown-features/react#importing-markdown) in Docusaurus. This would make it so we have one source of truth that we can reuse and maintain.

content/docs/quickstart.mdx
ZPain8464: I think it makes most sense to include it at the first mention of the Verify app, so earlier in the page. Will do.

content/docs/quickstart.mdx
ZPain8464: Well that was the original intention of the Fundamentals courses. This links to the Routes course first, so they can learn to build a route once they have Zero set up. The Zero Fundamentals course does not re-iterate the Quickstart; it instructs users to go back to the Quickstart and complete it first before continuing with Fundamentals.

PR #1330, opened 2024-03-26
Clusters Concepts page

ZPain8464:
This PR adds a Clusters Concepts page, which covers all the relevant details about a cluster that users might want to know.


content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Pomerium Zero generates and assigns a unique **Cluster Domain** to your cluster. Your cluster domain contains your randomly generated **Cluster Name** and the `pomerium.app` subdomain.
+
+It follows this format: `<CLUSTER-NAME-XXXX>.pomerium.app`
+
+For example: `voracious-ape-1578.pomerium.app`
+
+#### Change cluster Name
+
+If you don't like your randomly generated cluster name, you can change it manually in your cluster settings.
```
kenjenkins, line 65: To me this almost reads like you can change the starter domain by changing the cluster name, which I don't think is the case.

I wonder if this would be a good place to instead link to a guide for adding a custom domain.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,185 @@
...
+         - pomerium-cache:/var/cache
+
+   volumes:
+     pomerium-cache:
+   ```
+
+1. Run `docker compose up -d`
+
+</TabItem>
+<TabItem value="kubernetes" label="Kubernetes">
```
ZPain8464, line 102: Still need to confirm these steps are accurate for K8s.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,185 @@
...
+If you want to change the cluster name, you can update it in your cluster settings:
+
+1. From the navigation bar dropdown, select **Manage Clusters**
+1. Under **Actions**, select the **edit** icon
+1. In the **Cluster Name** field, enter your preferred cluster name
+1. Save your changes and apply the changeset
+
+![Viewing the new cluster name in the Zero Console](./img/clusters/cluster-name-table.png)
+
+#### Detected and Override IP address
```
ZPain8464, line 144: I personally think we should include this section here. At a minimum, we should cover these steps in Troubleshooting, or cross-link these pages.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+# Clusters in Pomerium Zero
+
+This document describes Clusters in Pomerium Zero.
+
+TODO: include a video that walks through Clusters
+
+## Clusters overview
+
+A **Cluster** represents a single Pomerium deployment managed by Pomerium Zero, our hosted control plane. Pomerium Zero enables you to configure cluster settings for your personal account or organization.
```
desimone, line 21: You can think of a Pomerium cluster as a completely distinct deployment of Pomerium, and allows you to effectively shard you deployments for performance, governance, security, or management purposes. 

I think we should probably discuss common reasons people would want to use a cluster.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,103 @@
...
+- **Cluster domain**
+- **Cluster settings**
+- **Routes and Policies**
+- **Wildcard certificates**
+
+### Cluster architecture
+
+TODO: It would be ideal to get an architecture diagram here.
+
+### Cluster identity
```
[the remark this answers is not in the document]
desimone: +1

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+- **Cluster identity**
+- **Cluster domain**
+- **Cluster settings**
+- **Routes and Policies**
+- **Wildcard certificates**
+
+### Cluster architecture
+
+TODO: It would be ideal to get an architecture diagram here.
```
desimone, line 33: An entity diagram would make the relationship more obvious.

ZPain8464:
I could use some input on the "Why clusters" section. It's pretty thin. The other TODOs would be to add a video, clusters architecture, and links for pending pages.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+#### All-in-one mode deployment model
+
+Pomerium consists of 4 services:
+
+- Proxy service
+- Authenticate service
+- Authorize service
+- Databroker service
+
+A cluster deploys these services in all-in-one mode, meaning the configuration for each service is managed from a single cluster instance. All inter-component communications are internally connected, with the exception of the Databroker service, which is an external dependency.
```
wasaga, line 40: I'm not sure this falls into "use case" as we do not really support other ways of deploying zero. 

Maybe we can mention (in a different section) that pomerium zero clusters are deployed as "all in one" and reference to https://www.pomerium.com/docs/internals/configuration#all-in-one-vs-split-service-mode

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+#### Multiple cluster instances
+
+You can deploy multiple clusters from a single cluster identity (see [**Cluster identity**](#cluster-identity) below for more information). The ability to deploy multiple cluster instances means you can manage and run completely separate configurations for workloads that may otherwise be incompatible in a single cluster instance.
```
wasaga, line 44: This is incorrect: you cannot deploy multiple clusters from a single cluster identity. 
you may, however, deploy multiple cluster _replicas_ from a single cluster identity. They, however, would share all the cluster configuration. 

> you can manage and run completely separate configurations for workloads that may otherwise be incompatible in a single cluster instance.

what is described is multiple clusters, not multiple cluster instances.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+### Cluster identity
+
+When you create an account in Pomerium Zero, a new **Cluster Token** is generated and assigned to your personal account. This cluster token, which is provided to you during onboarding, is required to deploy your cluster instance.
```
wasaga, line 48: While there is, indeed, an underlying cluster identity in zero, this is implementation detail and we do not directly expose it anywhere. 

Currently, in the UI we only talk about **Cluster Token**. 

image

```suggestion
When you create a cluster in Pomerium Zero, a **Cluster Token** is generated that allows your Pomerium instance to connect to the Pomerium Zero.
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+You can generate a new cluster token at any time. See the [**Rotate cluster identity**](#rotate-cluster-identity) steps below for more information.
+
+#### Rotate cluster identity
```
wasaga, line 52: While cluster identity in fact exists, this is implementation detail that is never exposed to the end user directly. I suggest we only mention **Cluster Token**, as this is the only thing end users see. 

```suggestion
#### Rotate cluster token
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+A cluster token does not have a set expiration time. However, you may want to rotate the cluster identity in the Zero Console for security reasons. For example:
```
wasaga, line 54:
```suggestion
A cluster token does not have a set expiration time. However, you may want to rotate the cluster token in the Zero Console for security reasons. For example:
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+- A malicious insider could use the token to view the cluster's configuration, which may contain sensitive data
```
wasaga, line 56:
```suggestion
- A cluster token secret was accidentally exposed, or
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+- You may lose track of the existing token, and want a new one
+
+:::caution
+
+Rotating the cluster identity will invalidate the existing cluster token.
+
+:::
+
+To rotate your cluster identity in the Zero Console:
```
wasaga, line 65:
```suggestion
To rotate your cluster token in the Zero Console:
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+1. From the navigation bar dropdown, select **Manage Clusters**
+1. Under **Actions**, select the **edit** icon
+1. Select the **Domains** tab
+1. In the **Cluster Identity** field, select **Rotate Identity**
```
wasaga, line 70: the described UI does not exist. it is currently in the _Manage > Settings > Advanced_ and is called **Rotate Token*

image

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+1. Copy the token value and store it somewhere safe ![Rotating a cluster identity in the Zero Console](./img/clusters/pz-rotate-cluster-identity.gif)
+
+After rotating the cluster identity, you must pass it into your configuration and run Pomerium. See the steps below for your deployment environment.
```
wasaga, line 73:
```suggestion
After rotating the cluster token, you must pass it into your configuration and run Pomerium. See the steps below for your deployment environment.
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+- Behind a firewall
+- In a Kubernetes server where the ingress load balancer IP address doesn't match the egress IP address
+
+In such cases, you may see a `DNS_PROBE_FINISHED_NXDOMAIN` error in your browser when you access a route. As a workaround, we've provided the **Override IP Address** field, which allows you to manually set the IP address so that Pomerium Zero can connect to your cluster.
+
+![Manually setting the cluster IP address in the Zero Console](./img/clusters/cluster-ip-address-settings.png)
+
+### Cluster settings
+
+Each cluster is assigned its own **Cluster settings** object, meaning the available settings within a cluster apply only to that cluster instance. This means changes applied to one cluster instance won't affect another cluster deployed with the same token.
```
wasaga, line 180:
```suggestion
Each cluster is assigned its own **Cluster settings** object, meaning the available settings within a cluster apply only to that cluster instance. This means changes applied to one cluster instance won't affect another cluster deployed with the same token.
```

this is incorrect. all clusters deployed with the same token share all the configuration (settings, routes, certs, etc).

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+### Routes and Policies
+
+**Routes** and **Policies** defined in a cluster are scoped only to that cluster, and are not available in other cluster instances.
```
wasaga, line 184:
```suggestion
**Routes** and **Policies** defined in a cluster are scoped only to that cluster, and are not available in other clusters.
```

wasaga (changes requested):
there is some confusion between different clusters and cluster instances. I tried to clarify that in the review comments.

additionally, I suggest we only use term **cluster token** and not use the term **cluster identity**, which is more like an implementation detail.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,241 @@
...
+
+## Cluster benefits
+
+### Simplified configuration
+
+Cluster configuration is managed entirely in the Zero Console. As you apply configuration changes in the Zero Console, your cluster fetches and applies these changes locally.
+
+This model ensures that:
+
+- Your cluster always stays in sync with the Zero Console as changes are applied
```
desimone, line 49:
```suggestion
- Your cluster remains synchronized with the Zero Console as changes are applied
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,241 @@
...
+- You don't have to manage cluster configuration files or environment variables
+- The Zero Console persists your cluster's last-known configuration
+
+### Data privacy
+
+Pomerium doesn't monitor your organization's traffic or collect private user data. Pomerium _does_ store deployment and configuration details that are necessary to control access to your services, such as your routes, policies, and TLS certificates.
```
desimone, line 55:
```suggestion
Pomerium cannot and does not monitor your organization's traffic or collect private user data. Pomerium does, however,  needs to store configuration details required to maintain the hosted control plane. Pomerium only stores deployment and configuration details needed to control access to your services, such as routes, policies, and TLS certificates and encrypts sensitive entities when appropriate.
```

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,241 @@
...
+#### Rotate cluster token
+
+A cluster token does not have a set expiration time. However, you may want to rotate the cluster token in the Zero Console for security reasons. For example:
+
+- A cluster token was accidentally exposed
+- You may lose track of the existing token, and want a new one
+
+:::caution
+
+Rotating the cluster token will invalidate the existing cluster token.
```
desimone, line 80:
```suggestion
Rotating the cluster token will invalidate the current token.
```

desimone (commented):
s/Zero Console/Pomerium Zero/h

ZPain8464:
@wasaga if I could get a review sometime this afternoon, that'd be great.

content/docs/concepts/clusters.mdx
```diff
@@ -0,0 +1,274 @@
...
+
+A cluster model reduces the complexity of managing Pomerium: All of a cluster's configuration, including settings, routes, policies, and certificates, are managed from a central control plane hosted in the Pomerium Zero cloud. This allows you to manage multiple isolated Pomerium Core deployments, each with its own configuration.
+
+Configuration changes applied in the hosted control plane are communicated over a streaming connection to your local cluster. Each cluster is connected to its own storage backend, which synchronizes state across replicas.
+
+    ![A diagram depicting how clusters communicate with the Zero Console](./img/clusters/clusters-architecture.png)
+
+Each replica runs Pomerium Core in an all-in-one mode, meaning the services that comprise Core are deployed together and share session state. Similarly, replicas communicate with each other to synchronize state across a cluster.
+
+The interconnectivity between replicas enables you to horizontally scale your cluster to adapt to the purposes of your use case, be it performance, governance, security, or management related.
```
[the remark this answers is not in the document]
ZPain8464: I updated the text. Thanks for the in-depth review/s, @wasaga.

PR #1362, opened 2024-04-16
Adds policy video to ppl page

ZPain8464:
Adds the policies video to the PPL page.


ZPain8464:
@desimone can I get this reviewed?

desimone (approved):
content LGTM but video quality worse than loom direct? What would a loom embed look like? I think we should still post youtube for discoverability / domain recognition when sharing but the quality difference is pretty big.

PR #1368, opened 2024-04-23
Adds prompt to change email in config.yaml like enterprise quickstart

gaurdro:
I noticed the note for updated the email in config.yaml was omitted from the non-enterprise quickstart, which would land a user at 403 forbidden instead of the desired page.


ZPain8464 (approved):
Thanks for adding this :)

PR #1374, opened 2024-04-25
Update desktop docs

### LINKED ISSUE #1366: Update Desktop Client documentation around client certificate options
Page: https://www.pomerium.com/docs/capabilities/tcp/client#add-a-connection

Desktop Client v0.23.0 adds a new "Search OS certificate store" option for client certificates, on Windows and macOS.

## What's the resolution?

There's a brief mention of client certificates on this page:
> **Client Certificate & Certificate Key File or Text**: For routes that require client certificates for [mTLS](https://www.pomerium.com/docs/concepts/mutual-auth), you can provide the certificate and key file to the Pomerium Desktop client.

Let's also mention the new option here.

While we're here, we might also want to update the Desktop Client [screenshot](https://github.com/pomerium/documentation/blob/74308e04e34fe5dae3255b100e4badf15257f078/content/docs/capabilities/tcp/examples/img/desktop/demo-new-connection.png) as well, as it's a few versions out of date.

ZPain8464:
@kenjenkins please let me know if you'd prefer different screenshots or more details for this blurb. 

Resolves #1366


ZPain8464:
@kenjenkins just realized this Pr includes a Zero image. We should probably remove it unless you don't think we need to backport this.

content/docs/capabilities/tcp/client.mdx
```diff
@@ -22,21 +22,32 @@ import LongLivedConnections from '@site/content/docs/admonitions/_long-lived-con
...
 # Pomerium Desktop and CLI Clients
 
-Pomerium is capable of creating secure connections to services like SSH, Redis, and more by creating a TCP tunnel to the service with a local client. This article describes configuring a route to accept TCP connections, and using either the CLI or GUI client to connect to it.
+This document describes how to use Pomerium's Desktop and CLI clients to connect to TCP routes in Pomerium.
 
-## Create a TCP Route
+:::note TCP Terminology
 
-1. Specify this new Route as a TCP Route by prefixing `tcp+` in the **From** field, along with a port suffix.
+In this document, a "TCP route" refers to a route configured to accept TCP connections.
```
kenjenkins, line 29: I think this is a good idea — it's natural for a reader to have the question "what's a TCP route" here.

I think we could make this even clearer by comparing with "regular" http/https routes.

What do you think of something like this?

```markdown
:::info What's a TCP route?

A TCP route is for use with upstream services that don't speak HTTP — for example
SSH, Redis, or MySQL. With a TCP route, the entire connection is proxied to the
upstream service, rather than each individual request separately.

In Pomerium, TCP routes are denoted with a `tcp+` prefix in the route's **From** URL.
```

content/docs/capabilities/tcp/client.mdx
```diff
@@ -22,21 +22,32 @@ import LongLivedConnections from '@site/content/docs/admonitions/_long-lived-con
...
 
-   The port is not used to connect to the Pomerium Proxy service from the internet; this will always be port 443 (unless otherwise defined in `config.yaml`). Rather, the port defined in **From** is part of the mapping to the individual route. In this way, you can create multiple routes that share a DNS entry, differentiated by the port to determine which route they use.
+:::
 
-   For example, suppose we have a server called `augur` running behind Pomerium that has a MySQL server and also listens for SSH connections. We can create routes for `tcp+https://augur.example.com:22` and `tcp+https://augur.example.com:3306`.
+## Create a TCP route
 
-1. The **To** field uses `tcp://` as a protocol, and specifies the address and port the service listens on.
+1. Specify this new Route as a TCP Route by prefixing `tcp+` in the **From** field, along with a port suffix
```
kenjenkins, line 35: Like below, I think "port number" is more standard.

```suggestion
1. Specify this new Route as a TCP Route by prefixing `tcp+` in the **From** field, along with a port number.
```

content/docs/capabilities/tcp/client.mdx
```diff
@@ -22,21 +22,32 @@ import LongLivedConnections from '@site/content/docs/admonitions/_long-lived-con
...
+1. The **To** field uses `tcp://` as a protocol, and specifies the address and port the service listens on
```
kenjenkins, line 36: While we're here, I think it would make sense to rephrase this step as an instruction, for consistency with step 1.

```suggestion
1. Set the **To** field using `tcp://` as the scheme, along with the address and port that the upstream service is listening on.
```

content/docs/capabilities/tcp/client.mdx
```diff
@@ -22,21 +22,32 @@ import LongLivedConnections from '@site/content/docs/admonitions/_long-lived-con
...
 
-The example below demonstrates a route to the SSH service on the host running the Pomerium Core or Pomerium Enterprise service:
+The example below demonstrates a route to an SSH service on the host running Pomerium:
 
 <Tabs>
+<TabItem value="Pomerium Zero" label="Pomerium Zero">
```
kenjenkins, line 41: I like the Zero screenshot, but I think it might be better to add this in a separate PR, so we can go ahead and backport this one.

content/docs/capabilities/tcp/client.mdx
```diff
@@ -49,116 +60,59 @@ The example below demonstrates a route to the SSH service on the host running th
...
-In most cases, you only need to specify the port (ex: `:2222`), and the client will listen on all available local addresses.
+- **Name**: A local name for the route
+- **Destination**: Matches the [From](/docs/reference/routes/from) value of the route, without the protocol. Always include the port specified in the route, and do not include the `https://` protocol.
+- **Local Address**: The local address and port number from which to access the service locally. If left blank, the client will choose a random port to listen to on the loopback address.
+- **Tags**: Customizable tags to sort and organize TCP routes
 
-**Tags**: Use tags to sort and organize your TCP routes.
+![Adding a new connection in the Pomerium Desktop client](./examples/img/desktop/desktop-new-connection.png)
 
 :::note Long-lived connections behavior
```
kenjenkins, line 95: This admonition feels out of place to me. I think it might make sense to remove it, so this section can focus on the "how-to".

kenjenkins (approved):
Thanks for updating this!

content/docs/capabilities/tcp/client.mdx
ZPain8464: Nice catch, and I agree.

content/docs/capabilities/tcp/client.mdx
```diff
@@ -49,116 +60,59 @@ The example below demonstrates a route to the SSH service on the host running th
...
-If you want Pomerium Desktop to start automatically when you log in to your computer, follow the steps below for your operating system.
-
-<Tabs>
-<TabItem value="windows" label="Windows">
-
-#### Autostart for all users
+:::note
 
-Copy the shortcut for the Pomerium Desktop app into `C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Startup`.
+The port suffix appended to the **From** route is not used to connect to the Pomerium Proxy service from the internet; this will always be port 443 (unless otherwise defined in the configuration file).
```
kenjenkins, line 68: Sorry, I think I deleted a pending comment somehow. I think this could be made a little clearer, along the lines of:

```markdown
:::tip

The port number in the route **From** URL is not used in the initial connection
to Pomerium itself. This connection will still use port 443, unless you use a
bastion host (see [Advanced configuration](#advanced-configuration) below).
```

PR #1385, opened 2024-05-07
Updates ppl example

### LINKED ISSUE #1383: Example PPL policy is potentially confusing
Page: https://www.pomerium.com/docs/capabilities/ppl#at-a-glance

There are a couple things about this policy that may be confusing or misleading:

1. The policy includes `user` criteria with email addresses.

   The `user` criterion matches the user ID, not email address. Depending on the IdP these _might_ be the same, but in general they are not.

2. The combination of criteria in the `allow` rule appears to be redundant.

   If you already require a specific user ID with a `user` criterion, I don't know why you would also include a `domain` criterion.

## What's the resolution?

Let's see if we can come up with a clearer example policy. Let's make sure any email address matching uses the `email` criterion.

ZPain8464:
Resolves #1383 

@kenjenkins I only made the changes we discussed. Please feel free to add any other suggestions while we're here.

### DIFF
--- a/content/docs/capabilities/ppl.mdx
+++ b/content/docs/capabilities/ppl.mdx
@@ -34,8 +34,8 @@ Each PPL policy has at the top level a set of `allow` or `deny` actions, with a 
 ```yaml
 allow:
   and:
-    - email:
-        is: user1@example.com
+    - domain:
+        is: example.com
 deny:
   or:
     - email:
@@ -44,9 +44,9 @@ deny:
         is: user3@example.com
 ```
 
-This policy grants access only if a user's email address matches `user1@example.com`.
+This policy grants access only if the domain portion of a user's email address matches the specified value, `example.com`.
 
-It will deny `user2` **or** `user3`.
+It will deny access to users with a `user2@example.com` **or** `user3@example.com` email address.
 
 ## Rules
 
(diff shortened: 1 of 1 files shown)


content/docs/capabilities/ppl.mdx
```diff
@@ -34,19 +34,19 @@ Each PPL policy has at the top level a set of `allow` or `deny` actions, with a
 ```yaml
 allow:
   and:
-    - domain:
-        is: example.com
-    - user:
+    - email:
```
kenjenkins, line 37: If we're going to have just one condition in this block I'd recommend keeping the `domain` condition rather than `email`. I think that makes a little more sense as a policy: "allow everyone from this domain except for these two specific users"; otherwise it's not clear why you would need a `deny` block at all.

content/docs/capabilities/ppl.mdx
```diff
@@ -158,7 +158,7 @@ Entries marked with `*` denote criteria that are only available in the [Enterpri
 | \* `record` | variable | Allows policies to be extended using data from [external data sources](/docs/integrations) |
 | `reject` | Anything. Typically `true`. | Always returns false. The opposite of `accept`. |
 | \* `time_of_day` | [Time of Day Matcher] | Returns true if the time of the request (for the current day) matches the constraints. |
-| `user` | [String Matcher] | Returns true if the logged-in user's id matches the given value. |
+| `user` | [String Matcher] | Returns `true` if the logged-in user's ID matches the supplied value. (The actual value of the user ID claim depends on how the identity provider sets this value.) |
```
kenjenkins, line 161: Thanks, I think this is probably helpful clarification.

kenjenkins (approved):
Thanks!

Issue #1381, opened 2024-04-30
Document port matching for matching routes


ZPain8464:
Related to this [PR](https://github.com/pomerium/pomerium/pull/5085). We need to document this new behavior for **From** URLs. 

- [x] Update From reference page - document what happens when a port is specified
- [x] Add a new Runtime Flags reference page to house all supported runtime flags

PR #1394, opened 2024-05-13
Updates routes for port mapping

ZPain8464:
Resolves https://github.com/pomerium/documentation/issues/1381.

@calebdoxsey  this is a WIP because I've added a Zero screenshot. For now, I'd like a review for the docs. i can remove the screenshot and tab for v0.26, and wait until we're ready to release Zero.

I also need to document the Kubernetes configuration for this.


content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,66 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
-**From** is the externally accessible URL for the proxied request.
-
-Specifying `tcp+https` for the scheme enables [TCP proxying](/docs/capabilities/tcp/) support for the route. You may map more than one port through the same hostname by specifying a different `:port` in the URL.
```
calebdoxsey, line 21: TCP routes were not changed.

content/docs/reference/runtime-flags.md
```diff
@@ -0,0 +1,80 @@
...
+https://www.example.com
+
+https://www.example.com:443
+
+https://www.example.com:8443
+
+https://www.example.com:18443
+```
+
+If you disable this runtime flag and _do not_ specify a port in the From URL, Pomerium will only match this route if the incoming request _does not_ specify a port, _or_ the request specifies port `:443`.
```
calebdoxsey, line 74: it's the port that the server listens on, which is by default `443`

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
...
+The **From** route is the externally accessible URL for a proxied HTTP request.
 
 Specifying `tcp+https` for the scheme enables [TCP proxying](/docs/capabilities/tcp/) support for the route. You may map more than one port through the same hostname by specifying a different `:port` in the URL.
 
 ## How to configure
 
-The `URL` must contain a **scheme** and **hostname** and cannot contain a path.
+The From URL must contain a **scheme** and **hostname**. It can't contain a path.
+
+### Port mapping behavior
```
kenjenkins, line 28: I think we might want to stick with the phrase "port matching" here. According to wikipedia ["port mapping"](https://en.wikipedia.org/wiki/Port_mapping) has a meaning in the context of NAT, which is something unrelated.

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
 
 ## Summary
 
-**From** is the externally accessible URL for the proxied request.
+The **From** route is the externally accessible URL for a proxied HTTP request.
```
kenjenkins, line 20: I don't think we should use the phrase "From route." A route must have both a from URL and a to URL. There aren't "from routes" or "to routes".

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
...
 Specifying `tcp+https` for the scheme enables [TCP proxying](/docs/capabilities/tcp/) support for the route. You may map more than one port through the same hostname by specifying a different `:port` in the URL.
 
 ## How to configure
 
-The `URL` must contain a **scheme** and **hostname** and cannot contain a path.
+The From URL must contain a **scheme** and **hostname**. It can't contain a path.
+
+### Port mapping behavior
+
+In v0.26, we've updated the default route matching behavior so that From routes defined _without_ a specific port in the route URL will match a request on any port.
```
kenjenkins, line 30: Likewise here, let's be precise with our terminology: "from route" doesn't make sense to me, and neither does "route URL" (there are two URLs, the From URL and the To URL).

I also think it might make sense to describe the behavior fully up front, and then proceed with the examples.

```suggestion
In v0.26, we've updated the route matching behavior. If you do not specify a port number in a route's From URL, the route will match incoming requests with any port number (or no port number). If you do specify a port number, that route will match only requests containing that specific port number.
```

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
...
+
+:::note
+
+You can disable this behavior with the [**Match Any Incoming Port**](/docs/reference/runtime-flags) runtime flag.
```
kenjenkins, line 34: I'd suggest moving this to the bottom of the section, so it doesn't interrupt the explanation + examples.

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
...
+If you specify the port in the From URL, Pomerium will only match incoming requests that specify this port:
+
+```yaml
+from: https://www.example.com:8443
+```
 
 :::note
 
-Only secure schemes (`https` and `tcp+https`) are supported.
+When defining a From route, you must use `https` or `tcp+https`. Pomerium only supports secure schemes.
```
kenjenkins, line 64: I'd suggest moving this note up, so it stays together with the sentence "The From URL must contain a scheme and hostname".

Or, it might make sense to remove the `:::note` callout and add this directly to that paragraph:

> The From URL must contain a **scheme** and **hostname**. It can't contain a path. The scheme must be `https` or `tcp+https`.

content/docs/reference/routes/from.mdx
```diff
@@ -16,21 +17,62 @@ import TabItem from '@theme/TabItem';
...
 
 :::
 
 <Tabs>
+<TabItem value="zero" label="Zero">
```
kenjenkins, line 69: I think if we keep this change in a separate PR it may make things easier when cutting the v0.26 docs site branch.

content/docs/reference/routes/from.mdx
ZPain8464: Done.

content/docs/reference/routes/from.mdx
ZPain8464: So a "route" consists of a From and To URL. I agree here. I thought it made sense to distinguish which one of these we're referring to. "From route" could be "From URL".

content/docs/reference/routes/from.mdx
ZPain8464: Done.

content/docs/reference/routes/from.mdx
ZPain8464: Agreed. I'll remove and add in a separate PR.

Issue #1389, opened 2024-05-09
add a header for legacy version docs that they are not for current version


wasaga:
<img width="822" alt="image" src="https://github.com/pomerium/documentation/assets/673131/4557611a-a399-411c-866e-141fe80c8b3f">

## What's incorrect or missing

The user may not necessarily notice that these are not current docs. 
We've got users citing legacy version documentation every week, even if they're just starting with pomerium. 

## What's the resolution?

Add a header that documentation is not for current version, and link to current.

desimone:
_great_ idea. +1

kenjenkins:
@desimone, do you have any context on how the older docs sites are hosted? Would we need to restore some of the Git history for the docs site deleted in https://github.com/pomerium/pomerium/pull/3420 in order to make updates to v0.10 through v0.17?

ZPain8464:
@kenjenkins and @desimone I went ahead and started implementing this, but to Ken's point, the available branches in `pomerium/documentation` go back to `0-17-0`. Beyond that, it's difficult to retroactively add the announcement banner.

ssveta7ak:
Implemented.

PR #1396, opened 2024-05-13
Adds v0.24 outdated docs notice

ZPain8464:
Resolves https://github.com/pomerium/documentation/issues/1389.

### DIFF
--- a/content/docs/quickstart.mdx
+++ b/content/docs/quickstart.mdx
@@ -51,7 +51,7 @@ Add the configuration below to `docker-compose.yaml`:
 
 <DockerCompose />
 
-**Run Docker Compose:**
+## Run Docker Compose
 
 ```bash
 docker compose up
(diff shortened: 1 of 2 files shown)


content/docs/quickstart.mdx
```diff
@@ -51,7 +51,7 @@ Add the configuration below to `docker-compose.yaml`:
 
 <DockerCompose />
 
-## Run Docker Compose
+**Run Docker Compose:**
```
kenjenkins, line 54: What's this change for?

kenjenkins (approved):
Looks great! I wish we could keep it at the top of the page even when scrolling down, but I don't see a configuration option for that or an easy way to do that with a CSS tweak.

content/docs/quickstart.mdx
ZPain8464: I may have done that by accident when resolving a merge conflict. It should be a an H2.

PR #1407, opened 2024-05-14
FIX: use correct config key for cookie expiration

aponert:
The here documented key cookie_expiration does not work (tested using 0.25.2), cookie_expire does work, which was documented in an older version of the docs. If the docs are meant to be correct, then maybe pomerium itself should be changed to use the updated syntax


ZPain8464 (approved):

ZPain8464:
@aponert thanks for the contribution here. Yes, this key is incorrect. I'm backporting this PR as well so that older versions are updated.

PR #1416, opened 2024-05-20
Removes releases < v0.20 

ZPain8464:
This PR removes Core and Enterprise changelogs and upgrade guides up to v0.20.  This PR only resolves the removal of older versions; it does not resolve the second requirement found in https://github.com/pomerium/internal/issues/1798. 

Related to https://github.com/pomerium/internal/issues/1798.


desimone:
Yeah, I don't think we need to _scrub_ the changelog here but rather the versions link list.

desimone (changes requested):
Yeah, I don't think we need to scrub the changelog here but rather the versions link list.

ZPain8464:
Ah I see, I'll update this PR to remove the archived versions links up to v0.20, and restore the changelogs and upgrade guides.

kenjenkins (approved):

ZPain8464:
@desimone looks like you need to approve the changes to merge.

desimone (approved):

PR #1424, opened 2024-05-21
adds zero sidebar

ZPain8464:
I closed https://github.com/pomerium/documentation/pull/1388, as it was a draft and I kept adding changes.

Thoughts here:
- The "What is Pomerium Zero?" page is incomplete, but I think in its current state, we can add to it. Screenshots, possibly a video, would improve it immensely. I'd like to merge this as soon as possible, as several other pending Zero docs, including an Upgrading page and a Health Checks page, will live in this sidebar.
- @wasaga can you please review the "Health Checks" doc? 
- @kenjenkins do you mind reviewing the upgrade guide?


content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+    latest zero version,
+  ]
+pagination_prev: null
+pagination_next: null
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Upgrade Pomerium Zero
```
kenjenkins, line 21: I think we should be careful with our terminology. What does "Pomerium Zero" refer to? I'd expect this to mean "our managed control plane", in which case I'm not sure it makes sense to talk about "upgrading Pomerium Zero."

How do you think we should talk about this?

I'd suggest talking about "upgrading Pomerium" or "upgrading your Pomerium Zero deployment."

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+  image: pomerium/pomerium:vX.Y.Z
+```
+
+1. Run the following command:
+
+```bash
+$ docker compose up -d
+```
+
+Docker should automatically pull the new image of Pomerium Zero before running the container. If for some reason it doesn't pull the image, you can manually run:
```
kenjenkins, line 51: Here as well, I think it's confusing to talk about an "image of Pomerium Zero" because this is the regular pomerium/pomerium docker image, not something specific to Pomerium Zero.

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+:::
+
+## How to upgrade Pomerium Zero
+
+<Tabs>
+<TabItem label="Docker" value="docker">
+
+To update Pomerium Zero in Docker:
+
+1. In your `docker-compose.yaml` file, update the Pomerium Zero image tag to specify the [latest tagged release](https://github.com/pomerium/pomerium/tags):
```
kenjenkins, line 38: I think the onboarding instructions say to save the configuration just as `compose.yaml`? Might want to double check the file name for consistency.

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Upgrade Pomerium Zero
+
+This page shows you how to upgrade your cluster to use the latest version of Pomerium Zero. Refer to the steps for your deployment environment.
+
+:::note
+
+Review the latest [**Core Upgrade guide**](/docs/deploy/core/upgrading) for release details before you upgrade Pomerium Zero.
```
kenjenkins, line 27: If we're managing configuration on behalf of users, I think we should strive to handle any backwards incompatible changes ourselves. I'd recommend we not mention the upgrade guide here.

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+</TabItem>
+<TabItem label="Linux" value="linux">
+
+To update Pomerium Zero in Debian-based Linux systems:
+
+1. Check for new package updates:
+
+```bash
+$ sudo apt update
```
kenjenkins, line 65: (optional) might be easier to combine this into one command:
```
$ sudo apt update && sudo apt install pomerium
```

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+```
+
+1. Install the latest version of Pomerium Zero:
+
+```bash
+$ sudo apt --only-upgrade install pomerium
```
kenjenkins, line 71: I don't know that we need the `--only-upgrade` option?

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+1. Install the latest version of Pomerium Zero:
+
+```bash
+$ sudo yum update pomerium
+```
+
+Then, run Pomerium:
+
+```bash
+$ sudo systemctl start pomerium
```
kenjenkins, line 91: I don't think we need this step? I think systemd should automatically restart Pomerium after the update.

content/docs/zero/upgrading.mdx
```diff
@@ -0,0 +1,104 @@
...
+Docker should automatically pull the new image of Pomerium Zero before running the container. If for some reason it doesn't pull the image, you can manually run:
+
+```bash
+$ docker pull pomerium/pomerium:vX.Y.Z
+```
+
+</TabItem>
+<TabItem label="Linux" value="linux">
+
+To update Pomerium Zero in Debian-based Linux systems:
```
kenjenkins, line 60: I think we might want to add a warning here for anyone that used the Linux install script during the beta period (due to https://github.com/pomerium/pomerium-zero/issues/2001). I can work on this.

kenjenkins (commented):
Thanks for working on this!

content/docs/zero/health-checks.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+**What causes this error**: If you see this warning, please [contact support](https://discuss.pomerium.com/).
+
+## There's an issue with the network listener {#xds-listener}
+
+This health check makes sure the network listener is configured correctly. It runs every time the cluster configuration is updated.
+
+**What causes this error**: Typically, this error is caused when Pomerium's network listener can't bind to the cluster's port.
+
+**Steps to resolve**: If this is a port binding error, you may need to kill a process running on that port or specify a port to bind to.
```
wasaga, line 93: another frequent issue is lacking permissions binding to port. user may need to add `CAP_NET_BIND_SERVICE` capability to the process.

wasaga (commented):
reviewed health checks

content/docs/zero/upgrading.mdx
ZPain8464: I think you're correct, I'll remove.

content/docs/zero/upgrading.mdx
ZPain8464: @kenjenkins do you intend to add that notice in this PR or in a separate PR? Either way is fine.

content/docs/zero/upgrading.mdx
ZPain8464: I've never used it when upgrading in Linux, but I wasn't sure if this is a best practice I'm just not aware of. Happy to remove.

content/docs/zero/upgrading.mdx
kenjenkins: Here's one option. (Or alternatively, maybe this would be a good item to add to an FAQ or troubleshooting page, and then we could link to that from here?)

    <details>
    <summary>
    Note: if you installed Pomerium using the Linux install script during the Pomerium Zero beta, you will need to re-run the install script the first time you upgrade Pomerium. (Subsequent upgrades will not require this step.)
    </summary>
    
    First find your current cluster token: look for a line beginning with `Environment=POMERIUM_ZERO_TOKEN=` in the file /usr/lib/systemd/system/pomerium.service. Copy this token into the following command and run it:
    
    ```bash
    $ curl https://console.pomerium.app/install.bash | \
      env POMERIUM_ZERO_TOKEN=<your token here> bash -s install
    ```
    
    </details>

ZPain8464: I like the idea of linking to the Zero FAQ page from here. When it's merged, we can add this.

ZPain8464: @kenjenkins I added this as an entry to the troubleshooting page in https://github.com/pomerium/documentation/pull/1426. Feel free to review if you'd like, I'm holding off on merging for a bit.

PR #1434, opened 2024-05-28
WIP -- experimenting with navbar product selector

kenjenkins:
Docusaurus apparently does support custom properties on sidebar items, as long as they are nested under a 'customProps' object. (Individual pages can set these properties using the 'sidebar_custom_props' front matter key.)

Swizzle the DocSidebar component to add a wrapper that will filter the sidebar items depending on a 'products' custom property. Add a new custom navbar component to switch between products.

This doesn't work very well yet:
 - the currently-selected product isn't persisted anywhere
 - the dropdown menu doesn't dismiss when clicking an item
 - the active product is not highlighted in the dropdown menu


ZPain8464:
@kenjenkins I appreciate the help here. If we wanted to move forward with this approach, what would be next? I have a few items in mind:
- Make "Zero" the default product selection (this would enforce the "Zero as happy path" approach)
- Configure each product view so that the relevant "Quickstart" is the top-level item
- Refactor "Deploy" sidebar slice so that product-specific pages aren't so nested. For example, `/docs/deploy/k8s/ingress` might be `/docs/deploy/ingress` 
- Make it so Zero-specific docs, like Clusters and Custom Domains, only appear in Capabilities or Concepts if that product is selected

If this is the best path forward, I'd be happy to work on this.

PR #1443, opened 2024-06-07
Replaces loom iframes with  wistia inline iframes

ZPain8464:
This PR adds the inline iframes from Wistia for the Routing, Policies, and Custom Domains videos. @nhayfield I had to manually add the `iframe` parameter to the URL. It seems to work, but I wanted to get your eyes on it to make sure in case I missed something.


nhayfield (approved):
Code looks good.

Issue #1453, opened 2024-06-14
The old Twitter icon is displayed in the footer instead of the new X icon.


ssveta7ak:
https://www.pomerium.com/docs

## What's incorrect or missing
The old Twitter icon is displayed in the footer instead of the new X icon.
![Screenshot from 2024-06-14 12-11-21](https://github.com/pomerium/documentation/assets/37595288/929f54b0-84ee-47e0-8ee4-bc5242260442)

ssveta7ak:
Fixed.

PR #1454, opened 2024-06-14
update X/Twitter logo

kenjenkins:
Update font-awesome link to a newer version that includes a newer X/Twitter logo. Something changed about the way the CSS is structured, such that loading just the brands.min.css file is not sufficient to apply the proper font-family setting, so tweak custom.css to fix this.

Related issue:
 - #1453


ZPain8464 (approved):
Thanks for updating this, @kenjenkins. Looks good!

PR #1481, opened 2024-06-27
reference: add custom-domains link path

kenjenkins:
Add 'path' property to the custom-domains entry in reference.json, now that we have a published Capabilities page for custom domains in Zero.


ZPain8464 (approved):
Thanks for updating this, @kenjenkins.

Issue #1469, opened 2024-06-24
Add redirects for all pages moved in #1439


kenjenkins:
The updates to [_redirects](https://github.com/pomerium/documentation/pull/1439/files#diff-16b6401519950c5f36e2548e217df9174771c97bbfb0c8ead59906fe108d28fe) in https://github.com/pomerium/documentation/pull/1439 appear to be incomplete. Every moved page should have a corresponding redirect.

kenjenkins:
I think this incantation will show all the pages that were deleted or moved:

```console
% git show 1d26b1be81bf909e857f86acfc3be7f02ce6dce4 --diff-filter=D --no-renames --name-only | sed -nr 's/^content(.*)\.mdx?$/\1/p' 
/docs/concepts/certificates
/docs/deploy/clients
/docs/deploy/clients/pomerium-cli
/docs/deploy/clients/pomerium-desktop
/docs/deploy/core
/docs/deploy/core/binary
/docs/deploy/core/changelog
/docs/deploy/core/from-source
/docs/deploy/core/upgrading
/docs/deploy/enterprise
/docs/deploy/enterprise/changelog
/docs/deploy/enterprise/configure
/docs/deploy/enterprise/install
/docs/deploy/enterprise/quickstart
/docs/deploy/enterprise/upgrading
/docs/deploy/k8s/configure
/docs/deploy/k8s/ingress
/docs/deploy/k8s/install
/docs/deploy/k8s/quickstart
/docs/deploy/k8s/reference
/docs/deploy/production-deployment
/docs/deploy/releases
/docs/deploying
/docs/guides/js-sdk
/docs/guides/local-oidc
/docs/guides/securing-tcp
/docs/zero/cluster-status
/docs/zero/upgrading
```

We can loop over the output to check for any paths that return an HTTP error code:
```console
% git show 1d26b1be81bf909e857f86acfc3be7f02ce6dce4 --diff-filter=D --no-renames --name-only | sed -nr 's/^content(.*)\.mdx?$/\1/p' | while read p; do (curl -sf https://www.pomerium.com$p > /dev/null || echo "Missing path $p"); done
Missing path /docs/deploy/clients
Missing path /docs/deploy/production-deployment
Missing path /docs/zero/upgrading
```

ZPain8464:
From DocsSync:
- `docs/deploy/clients`: The missing client page should point to the [clients.md](http://clients.md/) file.
[...]

PR #1482, opened 2024-07-01
Adds upgrading steps to install page 

ZPain8464:
This PR adds upgrading steps for Zero installations to the Zero Install at https://www.pomerium.com/docs/zero/install. 

It also adds a redirect for the removed Zero Upgrading page. 

@kralicky I requested a review from you here because I believe you pushed for the `manual` configuration steps in onboarding. Do you have suggestions for upgrade steps for manual installations? 

Resolves https://github.com/pomerium/internal/issues/1845

Related to https://github.com/pomerium/documentation/issues/1469

### DIFF
--- a/content/docs/zero/install.mdx
+++ b/content/docs/zero/install.mdx
@@ -104,7 +104,7 @@ Use the following values as necessary to install or update Pomerium in accordanc
 
 Pass the provided **Cluster Token** into the `POMERIUM_ZERO_TOKEN` environment variable.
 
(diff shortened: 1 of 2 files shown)


kralicky:
re: manual setup, other than updating the pomerium image tag I don't think there are any specific instructions we can give. It's probably reasonable to assume that users who have deployed pomerium in a custom configuration would know what needs to be done to perform upgrades to that configuration.

ZPain8464:
@kralicky thanks for the response here. Would you mind approving? I'll add the instruction to update the image tag for manual configs.

kralicky (approved):

Issue #1476, opened 2024-06-25
Request: combine "The Number of Trusted Hops" with "X-Forwarded-For" settings pages


kenjenkins:
Page: https://www.pomerium.com/docs/reference/the-number-of-trusted-hops

## What's incorrect or missing

I think this setting would be clearer if it were grouped with https://www.pomerium.com/docs/reference/x-forwarded-for-http-header. The `xff_num_trusted_hops` and `skip_xff_append` settings both affect the behavior of the `X-Forwarded-For` and `X-Forwarded-Proto` headers, so I think it would be good to document these together like we do with some of the other closely-related settings.

Also, the description could be clearer: "The Number of Trusted Hops setting reverse proxies in front of Pomerium" doesn't make sense.

## What's the resolution?

Let's combine these two into an "X-Forwarded-For Settings" page, and provide some more explanation and examples of how/when you might want to use these settings.

ZPain8464:
@kenjenkins I think this is a reasonable request. I'll do some research on the `X-Forwarded-For` and `X-Forwarded-Proto` headers to see how we can better communicate what these settings do. 

I may have some questions before pushing a PR. I'll notify you if I do.

PR #1488, opened 2024-07-01
Adds X-Forwarded-For settings page

ZPain8464:
This PR combines the X-Forwarded-For HTTP reference page and the XFF-Num-Trusted-Hops reference page into a single page called "X-Forwarded-For Settings". 

@kenjenkins I did a lot digging into the Envoy documentation to update the examples, but if they can be improved or you have suggestions for clearer examples, please update the PR!  This also goes for the setting descriptions themselves.

Resolves https://github.com/pomerium/documentation/issues/1476


content/docs/reference/x-forwarded-for-settings.mdx
```diff
@@ -0,0 +1,166 @@
...
+pagination_next: null
+toc_max_heading_level: 2
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# X-Forwarded-For Settings
+
+## X-Forwarded-For HTTP Headers
```
kenjenkins, line 23: I find this name pretty confusing. I'd suggest we standardize on "Skip XFF Append" for this (to match the config file key), but let me know what you think.

(I've also filed https://github.com/pomerium/internal/issues/1849 to reconsider how we label this setting in the Enterprise Console and in the Zero UI.)

kenjenkins (commented):
Thanks for working on this! I wanted to double check some of the details and in the process realized I was confused about how this all works myself.

I'll push some changes but please take another look and let me know if anything feels off.

content/docs/reference/x-forwarded-for-settings.mdx
ZPain8464: @kenjenkins I agree with you 100% on changing the setting name to match with the config file key.

kenjenkins (approved):

ZPain8464:
@kenjenkins as a follow-up, we need to make sure the Help Docs files in Console and Zero match the new setting names. I'll tag @nhayfield so he's aware.

PR #1494, opened 2024-07-02
Fixes broken anchors

ZPain8464:
After upgrading to Docusaurus v3.4, it now can list broken anchor links. This PR fixes a portion of these anchors throughout the docs. There are more to fix, but this PR handles a good amount of them. 

There are some anchors that are more difficult to determine the right link. I'll come back to these after consulting the team.


cmo-pomerium (approved):
Approved.

Issue #1502, opened 2024-07-08
404: https://www.pomerium.com/docs/releases/enterprise/about


ZPain8464:

ZPain8464:
Fixed. Redirects to https://www.pomerium.com/docs/enterprise

Issue #1499, opened 2024-07-08
404: https://www.pomerium.com/docs/deploy


ZPain8464:

ZPain8464:
Fixed. Redirects to https://www.pomerium.com/docs

Issue #1498, opened 2024-07-08
404: https://www.pomerium.com/docs/integrations


ZPain8464:
Fixed. Redirects to https://www.pomerium.com/docs/capabilities/integrations

PR #1503, opened 2024-07-08
Fixes a few 404 links from GSC

ZPain8464:
This PR fixes the following 404ing links pulled from GSC:

- [ ] #1498 
- [ ] #1499 
- [ ] #1500 
- [ ] #1501 
- [ ] #1502 

Related to https://github.com/pomerium/internal/issues/1841

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -153,19 +153,18 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 /enterprise/service-accounts.html /docs/capabilities/service-accounts
 /enterprise/prometheus.html /docs/capabilities/metrics
 /enterprise/prometheus /docs/capabilities/metrics
-/docs/enterprise/reference/ /docs/releases/enterprise/configure
-/docs/enterprise/reference/configure /docs/releases/enterprise/configure
-/docs/enterprise/reference/configure.html /docs/releases/enterprise/configure
-/docs/enterprise/reference/config /docs/releases/enterprise/configure
-/docs/enterprise/reference/config.html /docs/releases/enterprise/configure
-/enterprise/reference/configure.html /docs/releases/enterprise/configure
-/enterprise/reference/config.html /docs/releases/enterprise/configure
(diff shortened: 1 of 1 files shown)


ZPain8464:
@ssveta7ak we won't be able to test if these actually worked until they are merged.

static/_redirects
```diff
@@ -164,6 +164,7 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 # /docs/enterprise/about uses multiple redirects
 /docs/enterprise/about /docs/releases/enterprise
 /docs/releases/enterprise /docs/releases/enterprise/about
+/docs/releases/enterprise/about /docs/enterprise
```
kenjenkins, line 167: Would it make sense to simplify some of these redirects while you're here?

I think this line makes the redirect from /docs/enterprise → /docs/releases/enterprise/about on line 151 above redundant, so we could delete line 151.

And maybe a lot of this section could be simplified by removing `/releases` from the right hand side? (like on lines 156-162)

static/_redirects
```diff
@@ -245,6 +247,8 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 /docs/enterprise/reference/manage.html /docs/capabilities/authentication
 /docs/enterprise/reference/reports /docs/capabilities/reports
 /docs/enterprise/external-data /docs/integrations/
+/docs/integrations /docs/capabilities/integrations
+/docs/integrations /docs/capabilities/integrations
```
kenjenkins, line 251: Looks like this is a duplicate of the line above. Was there supposed to be something else here?

kenjenkins (approved):

static/_redirects
ZPain8464: Thanks, removed.

static/_redirects
ZPain8464: That works. I need to comb through this whole file to remove redundant links.

PR #1510, opened 2024-07-09
Adds 410 redirect to Production Deployment page

ZPain8464:
Based on Netlify's [redirect docs](https://docs.netlify.com/routing/redirects/redirect-options/#http-status-codes), I think we have to specify a page to redirect to, so I pointed this page to our docs homepage. I also added a `410` status code. 

Resolves https://github.com/pomerium/documentation/issues/1469


ZPain8464:
@kenjenkins after checking the URL with curl, it returns a 410 status code.

PR #1515, opened 2024-07-10
Update dependencies within constraints of package.json

gaurdro:
just the output of `yarn upgrade`


ZPain8464:
@nhayfield any here that we shouldn't update? Or any reason to not update the package.json file? @gaurdro this is still in Draft, do you plan on making it ready for review?

gaurdro:
I was waiting for the dependabot PR to go through before marking this one as ready to go.

nhayfield:
Should be fine as long as it works

ZPain8464 (approved):

PR #1518, opened 2024-07-11
Creates new service url page

ZPain8464:
This PR consolidates the following settings into a new Service URLs reference page:
- Authenticate Service URL
- Authenticate Internal Service URL 
- Authorize Service URL
- Authorize Internal Service URL

It attempts to rewrite the setting explanations, but I have a feeling it could use more clarity around the use cases of an internal service URL. 

Resolves https://github.com/pomerium/internal/issues/1850


content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,209 @@
...
+toc_max_heading_level: 2
+---
+
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import PomeriumAuthenticateServices from '@site/content/docs/admonitions/\_pomerium-authenticate-services.mdx';
+
+# Service URL Settings
+
+## Authenticate Service URL
+
+The **Authenticate Service URL** setting defines the externally accessible URL where Pomerium redirects end users (clients) to authenticate against an identity provider.
```
calebdoxsey, line 25: I think we should note here that this is only required if not using hosted authenticate.

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,209 @@
...
+
+If Pomerium is running in [split service mode](/docs/internals/configuration#all-in-one-vs-split-service-mode), each individual Pomerium service (except the Databroker service) requires the authenticate service URL in its configuration. These services include the Proxy, Authenticate, and Authorize services.
```
calebdoxsey, line 27: It actually is needed by the Databroker. The Databroker doesn't make a connection to it, but it ends up disabling session refresh because it thinks we're using hosted authenticate if undefined.

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,287 @@
+---
+title: Service URLs
```
kenjenkins, line 2: Should this be "Service URL Settings" to match the page heading below?

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,287 @@
...
+description: This page covers all of Pomerium's service URL settings, including authenticate, authorize, and databroker external and internal service URLs.
+keywords:
+  - authenticate service url
+  - authenticate internal service url
+  - authenticate_service_url
+  - authenticate_internal_service_url
+  - authorize service url
+  - authorize_service_url
+  - authorize internal service url
+  - authorize_internal_service_url
```
kenjenkins, line 13: Let's include databroker keywords here too, for consistency.

content/docs/reference/service-urls.md
kenjenkins: I think it could be helpful to add a little introduction (before any of the section headings) to spell out what these settings are for:
 - The authenticate service URL controls whether Pomerium will use the hosted authenticate service or not. If using your own IdP you will need this URL when configuring your IdP client (for the OAuth callback URL).
 - The other service URL settings are only for split service mode.

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+See the [Kubernetes - Global Configuration](/docs/k8s/configure) for more information.
+
+</TabItem>
+</Tabs>
+
+## Authenticate Internal Service URL
+
+The **Authenticate Internal Service URL** setting defines an internally accessible URL where Pomerium redirects end users (clients) to authenticate against an identity provider. Use this setting if Pomerium can't access an external service URL.
```
kenjenkins, line 84: "defines an internally accessible URL where Pomerium redirects end users" — I believe this is not correct. Users will only ever be redirected to the external authenticate service URL. The internal service URLs are for internal communication between Pomerium services.

I think the existing wording was more accurate: "Authenticate Internal Service URL is only required for [split-service mode](https://docs.pomerium.com/docs/internals/configuration#all-in-one-vs-split-service-mode) deployments where Pomerium can’t access the public [Authenticate Service URL](https://docs.pomerium.com/docs/reference/authenticate-service-url)."

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,287 @@
...
+
+You may include both an internal and external authenticate service URL in the same configuration. In this case, the internal authenticate service URL overrides the external service URL when Pomerium determines which TLS certificate and hostname the authenticate service will listen with when receiving incoming client requests.
+
+If Pomerium is running in [split-service mode](/docs/internals/configuration#all-in-one-vs-split-service-mode), each Pomerium service requires either an internal or external authenticate service URL in its configuration.
```
kenjenkins, line 88:
```suggestion
If an internal authenticate service URL is required, it should be configured on each Pomerium service.
```

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,287 @@
...
+See the [Kubernetes - Global Configuration](/docs/k8s/configure) for more information.
+
+</TabItem>
+</Tabs>
+
+## Authenticate Internal Service URL
+
+The **Authenticate Internal Service URL** setting defines an internally accessible URL where Pomerium redirects end users (clients) to authenticate against an identity provider. Use this setting if Pomerium can't access an external service URL.
+
+You may include both an internal and external authenticate service URL in the same configuration. In this case, the internal authenticate service URL overrides the external service URL when Pomerium determines which TLS certificate and hostname the authenticate service will listen with when receiving incoming client requests.
```
kenjenkins, line 86: I don't think the part about "overrides the external service URL when Pomerium determines which TLS certificate and hostname the authenticate service will listen" is accurate.

The internal authenticate service URL is used when another Pomerium service wants to communicate with the authenticate service.

content/docs/admonitions/_pomerium-authenticate-services.mdx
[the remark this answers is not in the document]
ZPain8464: Originally, we used this admonition on any setting that involved the authenticate service. This includes these settings we're consolidating (excluding the Databroker settings), and the following settings:
- [Authenticate Callback Path](https://www.pomerium.com/docs/reference/authenticate-callback-path) 
- [Identity Provider settings](https://www.pomerium.com/docs/reference/identity-provider-settings)

To better manage admonitions like this, I'm trying to consolidate these into a single source of truth where it makes sense. I'd need to update these other pages to import this admonition, though.

content/docs/admonitions/_pomerium-authenticate-services.mdx
```diff
@@ -0,0 +1,3 @@
+The Pomerium [**Hosted Authenticate Service**](/docs/capabilities/hosted-authenticate-service) runs by default in configurations that don't specify an identity provider or authenticate service URL.
```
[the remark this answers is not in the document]
ZPain8464: OK, makes sense. I'll update the text.

content/docs/reference/service-urls.md
ZPain8464: So, maybe this is something we should clarify for users? I was under the impression that "internal" meant the internal or private network, so basically Pomerium services could communicate with other non-Pomerium services with internal network addresses. Clearly that's wrong, and thanks for the clarification. 

To your second point in the comment above("The other service URL settings are only for split service mode."), would it be helpful to say that the internal settings are only for split-service mode deployments where Pomerium services can't communicate with the external URL?

content/docs/reference/service-urls.md
ZPain8464: So, I got that from the previous definition from v0.21, the version before these reference pages were updated. (See [Authenticate Internal Service URL](https://0-21-0.docs.pomerium.com/docs/reference/authenticate-internal-service-url).) 

Can we get clarity on what actually would happen if you included both URLs?

content/docs/reference/reference.json
```diff
@@ -26,9 +26,15 @@
     "type": "string",
     "services": ["proxy"]
   },
+  "service-urls": {
+    "id": "service-urls",
+    "path": "/service-urls",
+    "title": "Service URL Settings",
+    "description": "This page describes Pomerium's authenticate and authorize service URL settings."
```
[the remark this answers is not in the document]
ZPain8464: I've been creating references pages for these consolidated settings, but it seems unnecessary as it clutters up the reference table by adding more entries. Users search for a specific setting, not a page of related settings, which negates the need to write an entry for the page itself. 

I'll remove this entry and can remove other unnecessary entries in a separate PR.

content/docs/reference/service-urls.md
ZPain8464: It only affects how the title tag is rendered, but yes, these should be consistent. Updated.

content/docs/reference/service-urls.md
ZPain8464: I'm also not sure if adding the config file key and environment variable variants as keywords here is helpful for internal search, but that's the reasoning behind adding these.

content/docs/reference/service-urls.md
kenjenkins: This is my understanding: if you configure both Authenticate Service URL and Authenticate Internal Service URL, then the authenticate service will use the Authenticate Service URL for all the login / logout redirects, but other Pomerium services will use the Authenticate Internal Service URL when attempting to communicate with the authenticate service. (Although this may be less relevant since we restored the older authentication flow when not using the hosted authenticate service.)

It would probably be a good exercise to set up an environment with another proxy in front of Pomerium such that you would need to set the internal service URLs.

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import PomeriumAuthenticateServices from '@site/content/docs/admonitions/\_pomerium-authenticate-services.mdx';
+
+# Service URL Settings
+
+The [authenticate service URL](#authenticate-service-url) setting controls whether or not Pomerium will use the [hosted authenticate service](/docs/capabilities/hosted-authenticate-service). If you prefer to use your own [identity provider](/docs/identity-providers), you'll need this service URL when configuring your identity provider client's OAuth callback URL.
+
+## Authenticate Service URL
+
+The **Authenticate Service URL** setting defines the externally accessible URL where Pomerium redirects end users (clients) to authenticate against an identity provider.
```
kenjenkins, line 37:
```suggestion
The **Authenticate Service URL** setting defines the externally accessible URL where Pomerium redirects end users (clients) to authenticate against an identity provider.

If not set, Pomerium will use the [Hosted Authenticate Service](/docs/capabilities/hosted-authenticate-service).

If you prefer to use your own [identity provider](/docs/identity-providers), you'll need to set an authenticate service URL, and you will need this URL when configuring your identity provider client's OAuth callback URL.
```

content/docs/reference/service-urls.md
kenjenkins: I've added a few suggestions for tweaks here. I know I mentioned talking about the OAuth callback URL above but I'm thinking maybe it's better off here.

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+pagination_prev: null
+pagination_next: null
+toc_max_heading_level: 2
+---
+
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import PomeriumAuthenticateServices from '@site/content/docs/admonitions/\_pomerium-authenticate-services.mdx';
+
+# Service URL Settings
+
+The [authenticate service URL](#authenticate-service-url) setting controls whether or not Pomerium will use the [hosted authenticate service](/docs/capabilities/hosted-authenticate-service). If you prefer to use your own [identity provider](/docs/identity-providers), you'll need this service URL when configuring your identity provider client's OAuth callback URL.
```
kenjenkins, line 33:
```suggestion
Pomerium's service URL settings control how the different [Pomerium services](/docs/internals/architecture#component-level) communicate with each other.

The [authenticate service URL](#authenticate-service-url) setting also controls whether Pomerium will run its own authenticate service or use the [Hosted Authenticate Service](/docs/capabilities/hosted-authenticate-service).

The other service URL settings are needed only for split service deployments. When running in all-in-one mode (which is the recommended mode), these settings are not needed.
```

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+`authenticate_service_url` is a bootstrap configuration setting and is not configurable in the Console.
+
+</TabItem>
+<TabItem value="Kubernetes" label="Kubernetes">
+
+| **[Parameter name](/docs/k8s/reference#spec** | **Type** | **Usage**      |
+| :-------------------------------------------- | :------- | :------------- |
+| `authenticate.url`                            | `URL`    | \***required** |
+
+\* Excluding the `authenticate.url` defaults to the [hosted authenticated service](/docs/capabilities/hosted-authenticate-service).
```
kenjenkins, line 80:
```suggestion
| `authenticate.url`                            | `URL`    | **optional** |
```

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+### How to configure
+
+<Tabs>
+<TabItem value="Core" label="Core">
+
+| **Config file keys** | **Environment variables** | **Type** | **Usage** |
+| :-- | :-- | :-- | :-- |
+| `authenticate_service_url` | `AUTHENTICATE_SERVICE_URL` | `URL` | \***required** |
+
+\* Excluding the `authenticate_service_url` defaults to the [hosted authenticated service](/docs/capabilities/hosted-authenticate-service).
```
kenjenkins, line 56:
```suggestion
| **Config file keys** | **Environment variables** | **Type** | **Usage** |
| :-- | :-- | :-- | :-- |
| `authenticate_service_url` | `AUTHENTICATE_SERVICE_URL` | `URL` | **optional** |
```

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+\* Excluding the `authenticate.url` defaults to the [hosted authenticated service](/docs/capabilities/hosted-authenticate-service).
+
+### Examples
+
+```yaml
+authenticate:
+  url: https://authenticate.corp.example.com
+```
+
+See the [Kubernetes - Global Configuration](/docs/k8s/configure) for more information.
```
kenjenkins, line 89:
```suggestion
See the [Kubernetes - Global Configuration](/docs/k8s/configure#authenticate-endpoint) for more information.
```

content/docs/reference/service-urls.md
```diff
@@ -0,0 +1,297 @@
...
+
+</TabItem>
+</Tabs>
+
+## Authenticate Internal Service URL
+
+The **Authenticate Internal Service URL** setting is only required for split-service mode deployments where Pomerium can’t access the public Authenticate Service URL.
+
+If included, Authenticate Internal Service URL will override Authenticate Service URL.
```
kenjenkins, line 98:
```suggestion
If set, the Authenticate Internal Service URL will be used for communication between other Pomerium services and the authenticate service.
```

kenjenkins (approved):

PR #1525, opened 2024-07-16
update redirects to address some 404s from GSC

kenjenkins:
We found a few reference setting 404s from the Google Search Console report. I don't think we have any logs from Netlify (which might be able to help us infer whether any real users are hitting these), but I thought it would make sense to add redirects for these.

Related issue:
- https://github.com/pomerium/internal/issues/1841

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -88,7 +88,7 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 /guides/vs-code-server.html /docs/guides/code-server
 /guides/local-oidc.html /docs/guides/local-oidc
 /docs/guides/local-oidc /docs/identity-providers/oidc
-/docs/identity-providers/dex-freeipa /docs/identity-providers
+/docs/identity-providers/dex-freeipa /docs 410
 
 # Reference, capabilities, topics, concepts links
 /docs/reference/readme.html /docs/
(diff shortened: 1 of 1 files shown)


ZPain8464 (approved):

ZPain8464:
This PR is related to https://github.com/pomerium/internal/issues/1841

ZPain8464:
@kenjenkins the `production-deploy` redirect with a 410 status code appended does behave as expected in that the response includes a 410, but it doesn't redirect to the `/docs` page; it just behaves like a 404 page. I've removed the 410 status code to see if it redirects as expected.

SCR-20240717-jcyi

static/_redirects
```diff
@@ -46,16 +46,16 @@
 /guides/traefik-ingress.html https://0-20-0.docs.pomerium.com/docs/guides/traefik-ingress 301!
 /docs/reference/forward-auth https://0-20-0.docs.pomerium.com/docs/reference/forward-auth 301!
 https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.com/docs/guides
-/docs/reference/policy/allowed-domains https://0-14-0.docs.pomerium.com/reference
-/docs/reference/policy/allowed-idp-claims https://0-14-0.docs.pomerium.com/reference
-/docs/reference/policy/allowed-users https://0-14-0.docs.pomerium.com/reference
+/docs/reference/policy/allowed-domains https://0-14-0.docs.pomerium.com/reference/#allowed-domains
+/docs/reference/policy/allowed-idp-claims https://0-14-0.docs.pomerium.com/reference/#allowed-idp-claims
+/docs/reference/policy/allowed-groups https://0-14-0.docs.pomerium.com/reference/#allowed-groups
```
kenjenkins, line 51: I'm not sure whether it's worth adding this one (v0.14 was a long time ago), but I thought it might make sense to be consistent with the others. (And if we have them, I figured it made sense to add anchor tags so they link to the relevant section heading.)

static/_redirects
```diff
@@ -46,16 +46,16 @@
...
+/docs/reference/policy/allowed-users https://0-14-0.docs.pomerium.com/reference/#allowed-users
 /docs/reference/policy/policy-explanation https://0-14-0.docs.pomerium.com/reference
 /docs/reference/policy/policy-remediation https://0-14-0.docs.pomerium.com/reference
-/docs/reference/policy/policy https://0-14-0.docs.pomerium.com/reference
-
-# Link from v23 to v22
-/docs/reference/routes/headers#set-authorization-header https://0-22-0.docs.pomerium.com/reference/routes/headers#set-authorization-header
```
kenjenkins, line 57: I think this redirect will never trigger, because of the `#set-authorization-header` fragment on the source URL. (The fragment isn't part of the request sent to the server, so we can't issue a redirect based on it.) I figured we should just remove this line.

static/_redirects
ZPain8464: I agree for the sake of consistency, too. It can't hurt.

ZPain8464:
@kenjenkins I'd like to resolve this PR today if possible. I think this addresses all the important 404ing links for now.

PR #1528, opened 2024-07-18
Updates autocert settings reference page

ZPain8464:
This PR updates the Autocert settings page, and resolves https://github.com/pomerium/internal/issues/1855

### DIFF
--- a/content/docs/reference/autocert.mdx
+++ b/content/docs/reference/autocert.mdx
@@ -2,7 +2,7 @@
 id: autocert
 title: Autocert Settings
 description: |
-  Enable Autocert and Autocert settings to retrieve, manage, and renew public-facing TLS certificates from Let's Encrypt.
+  Autocert lets Pomerium manage public-facing TLS certificates automatically.
 keywords:
   - reference
   - Autocert
--- a/content/docs/reference/autocert.mdx
+++ b/content/docs/reference/autocert.mdx
@@ -104,7 +104,7 @@ AUTOCERT_CA=https://acme.zerossl.com/v2/DV90
 
 \* `string` must point to the directory path or URL of an S3 or GCS bucket.
 
-\*\* If you don't specify an Autocert directory, Pomerium will generate a directory by default based on your environment. See [Defaults](#autocert-directory-defaults) for more information.
+\*\* If you don't specify an Autocert directory, Pomerium will generate a directory by default based on your runtime environment. See [Defaults](#autocert-directory-defaults) for more information.
 
 #### Examples {#autocert-directory-examples}
 
@@ -272,7 +272,7 @@ AUTOCERT_TRUSTED_CA_FILE=/relative/file/location
 
 ## Autocert Use Staging {#autocert-use-staging}
 
-**Autocert Use Staging** setting allows you to use Let's Encrypt's [staging environment](https://letsencrypt.org/docs/staging-environment/), which has more generous [usage limits](https://letsencrypt.org/docs/staging-environment/) than the production environment. If you're testing or using Autocert in development, you should use the staging environment so that you don't exceed Let's Encrypt's rate limits.
+The **Autocert Use Staging** setting allows you to use Let's Encrypt's [staging environment](https://letsencrypt.org/docs/staging-environment/), which has more generous [usage limits](https://letsencrypt.org/docs/staging-environment/) than the production environment. If you're testing or using Autocert in development, you should use the staging environment so that you don't exceed Let's Encrypt's rate limits.
 
 ### How to configure {#autocert-use-staging-how-to-configure}
 
--- a/content/docs/reference/autocert.mdx
+++ b/content/docs/reference/autocert.mdx
@@ -2,10 +2,12 @@
 id: autocert
 title: Autocert Settings
 description: |
-  Enable Autocert and Autocert settings to retrieve, manage, and renew public facing TLS certificates from Let's Encrypt.
+  Enable Autocert and Autocert settings to retrieve, manage, and renew public-facing TLS certificates from Let's Encrypt.
 keywords:
   - reference
   - Autocert
+  - ACME
+  - acme
 pagination_prev: null
 pagination_next: null
 ---
--- a/content/docs/reference/autocert.mdx
+++ b/content/docs/reference/autocert.mdx
@@ -17,7 +17,7 @@ import TabItem from '@theme/TabItem';
 
 # Autocert Settings
 
-Pomerium’s Autocert settings provide a certificate issuance and management service in accordance with the [automated certificate management environment](https://letsencrypt.org/2019/03/11/acme-protocol-ietf-standard.html) (ACME) protocol.
+The Autocert settings configure Pomerium to manage public-facing TLS certificates automatically, by integrating with a certificate authority that supports the [ACME protocol](https://datatracker.ietf.org/doc/html/rfc8555), such as [Let's Encrypt](https://letsencrypt.org/).
 
 Autocert can be useful in situations where Pomerium may not be behind a TLS-terminating ingress controller, proxy, or load balancer that manages TLS certificates on behalf of your upstream services. Certificates provisioned by Autocert provide also [OCSP stapling](https://en.wikipedia.org/wiki/OCSP_stapling).
 
--- a/content/docs/reference/autocert.mdx
+++ b/content/docs/reference/autocert.mdx
@@ -8,6 +8,7 @@ keywords:
   - Autocert
   - ACME
   - acme
+  - OCSP stapling
 pagination_prev: null
 pagination_next: null
 ---
@@ -17,9 +18,13 @@ import TabItem from '@theme/TabItem';
 
 # Autocert Settings
 
-The Autocert settings configure Pomerium to manage public-facing TLS certificates automatically, by integrating with a certificate authority that supports the [ACME protocol](https://datatracker.ietf.org/doc/html/rfc8555), such as [Let's Encrypt](https://letsencrypt.org/).
+The Autocert settings configure Pomerium to manage public-facing TLS certificates automatically, by integrating with a certificate authority (CA) that supports the [ACME protocol](https://datatracker.ietf.org/doc/html/rfc8555), such as [Let's Encrypt](https://letsencrypt.org/).
 
 Autocert can be useful in situations where Pomerium may not be behind a TLS-terminating ingress controller, proxy, or load balancer that manages TLS certificates on behalf of your upstream services. Certificates provisioned by Autocert provide also [OCSP stapling](https://en.wikipedia.org/wiki/OCSP_stapling).
+
+You must use the [Autocert Directory](#autocert-directory) setting with Autocert to store and share certificate data between Pomerium services.
+
+Autocert can incorporate also certificates set manually in the Pomerium configuration. These certificates will take precedence over generated certificates when applicable to configured routes.
 
 :::note Autocert support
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
 
 # Autocert Settings
 
-This reference covers all of Pomerium's **Autocert Settings**:
+Pomerium’s Autocert settings provide a certificate issuance and management service in accordance with the [automated certificate management environment](https://letsencrypt.org/2019/03/11/acme-protocol-ietf-standard.html) (ACME) protocol.
```
kenjenkins, line 20: Let's be a little careful with the phrasing here. To me this reads like it's saying Pomerium is the certificate issuer, but that isn't the case.

I'm also wondering about the ACME protocol link — that URL looks like an announcement about the ACME protocol but I'm not sure if it gives the most useful context on what it is. I'm not sure if there's a nice summary explanation of this somewhere (the [wikipedia page](https://en.wikipedia.org/wiki/Automatic_Certificate_Management_Environment) introduction isn't super helpful either). There's always the official RFC: https://datatracker.ietf.org/doc/html/rfc8555.

What do you think of this?

```suggestion
The Autocert settings configure Pomerium to manage public-facing TLS certificates automatically, by integrating with a certificate authority that supports the [ACME protocol](https://datatracker.ietf.org/doc/html/rfc8555), such as [Let's Encrypt](https://letsencrypt.org/).
```

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
-The Autocert setting can be useful in situations where you may not have Pomerium behind a TLS terminating ingress or proxy that is already handling your public certificates on your behalf.
+- Pomerium Zero users should note that Pomerium automatically manages certificates for [starter domains](/docs/concepts/clusters#starter-domain) and [custom domains](/docs/capabilities/custom-domains).
 
-Autocert will incorporate certificates available in the system trust store and certificates set manually in the Pomerium configuration, and they will take precedence over generated certificates when applicable to configured routes.
+:::
+
+## Autocert {#autocert}
 
-Autocert will attempt `HTTP-01`and `TLS-ALPN-01` challenges. It does not support `DNS-01` challenges, required to generate wildcard certificates.
+The **Autocert** setting enables Pomerium to automatically retrieve, manage, and renew public-facing TLS certificates from [Let's Encrypt][letsencrypt] to secure Pomerium-managed route domains and the authenticate service (Autocert is not supported with the [Hosted Authenticate Service](/docs/capabilities/hosted-authenticate-service)).
```
kenjenkins, line 38: Where did the part about Hosted Authenticate come from?

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
 
-Kubernetes users should not use Autocert. See the [cert-manager's guide](https://cert-manager.io/docs/tutorials/acme/pomerium-ingress/) for more information.
+Autocert can incorporate also certificates available in the system trust store and certificates set manually in the Pomerium configuration. These certificates will take precedence over generated certificates when applicable to configured routes.
```
kenjenkins, line 40: I think I mentioned this in Slack, but I don't know that this is accurate. I think it would be unusual to configure a server application certificate via the system trust store.

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
+You must use the [Autocert Directory](#autocert-directory) setting with Autocert to store and share certificate data between Pomerium services.
 
-Autocert requires that port `443` be accessible from the internet in order to complete a [TLS-ALPN-01 challenge](https://letsencrypt.org/docs/challenge-types/#tls-alpn-01) or port `80` in order to complete an [HTTP-01 challenge](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) (https://letsencrypt.org/docs/challenge-types/#tls-alpn-01).
+:::caution Autocert usage limits
+
+By using Autocert, you agree to the [Let's Encrypt Subscriber Agreement](https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf). Let's Encrypt enforces [_strict_ usage limits](https://letsencrypt.org/docs/rate-limits/) per domain you should be aware of. Consider testing Autocert with [`autocert_use_staging`](/docs/reference/autocert) first.
```
kenjenkins, line 48: Looks like there's a newer version of the subscriber agreement: https://letsencrypt.org/documents/LE-SA-v1.4-April-3-2024.pdf (linked from https://letsencrypt.org/repository/#let-s-encrypt-subscriber-agreement).

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
-You must use [Autocert Directory](/docs/reference/autocert) with Autocert to store and share certificate data between services. (Note that Autocert also provides [OCSP stapling](https://en.wikipedia.org/wiki/OCSP_stapling).)
+- Kubernetes users should refer to the [Cert-Manager integration](/docs/k8s/ingress#cert-manager-integration) section for more information on configuring automated certificate management in the Pomerium Ingress Controller.
 
-The Autocert setting can be useful in situations where you may not have Pomerium behind a TLS terminating ingress or proxy that is already handling your public certificates on your behalf.
+- Pomerium Zero users should note that Pomerium automatically manages certificates for [starter domains](/docs/concepts/clusters#starter-domain) and [custom domains](/docs/capabilities/custom-domains).
 
-Autocert will incorporate certificates available in the system trust store and certificates set manually in the Pomerium configuration, and they will take precedence over generated certificates when applicable to configured routes.
+:::
+
+## Autocert {#autocert}
```
kenjenkins, line 36: I think a lot of the information in this section could be moved up into the overall summary.

High level comment: I think we should aim to structure this so we put the most important information first, and I'm not sure we achieve that so much of the information is tucked away under the setting headers and the headers are ordered alphabetically.

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
 
-Kubernetes users should not use Autocert. See the [cert-manager's guide](https://cert-manager.io/docs/tutorials/acme/pomerium-ingress/) for more information.
+Autocert can incorporate also certificates available in the system trust store and certificates set manually in the Pomerium configuration. These certificates will take precedence over generated certificates when applicable to configured routes.
 
-:::warning
+Before Autocert can provision certificates, it must complete either the [HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) or [TLS-ALPN-01](https://letsencrypt.org/docs/challenge-types/#tls-alpn-01) challenge. The HTTP-01 challenge requires internet access to port 80; the TLS-ALPN-01 challenge requires internet access to port 443. Autocert does not support `DNS-01` challenges, which is required to generate wildcard certificates.
```
kenjenkins, line 42: I think this could be clearer— "before" feels like the wrong word to describe this process. Maybe something more like this? "When requesting or renewing a certificate, Pomerium performs either the HTTP-01 or TLS-ALPN-01 challenge. The DNS-01 challenge is not supported."

The part about "requires internet access to port 80" also feels a little muddled. The important points are:
- using the HTTP-01 challenge requires setting up an HTTP redirect listener on port 80 — we can link to /docs/reference/http-redirect-address here
- using the TLS-ALPN-01 challenge requires that the Address setting uses port 443

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
 
-By using Autocert, you agree to the [Let's Encrypt Subscriber Agreement](https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf). There are [_strict_ usage limits](https://letsencrypt.org/docs/rate-limits/) per domain you should be aware of. Consider testing with [`autocert_use_staging`](/docs/reference/autocert) first.
+You must use the [Autocert Directory](#autocert-directory) setting with Autocert to store and share certificate data between Pomerium services.
```
kenjenkins, line 44: I think this is a good example of something that should be higher up on the page.

content/docs/reference/autocert.mdx
```diff
@@ -15,47 +17,42 @@ import TabItem from '@theme/TabItem';
...
 
-Autocert requires that port `443` be accessible from the internet in order to complete a [TLS-ALPN-01 challenge](https://letsencrypt.org/docs/challenge-types/#tls-alpn-01) or port `80` in order to complete an [HTTP-01 challenge](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) (https://letsencrypt.org/docs/challenge-types/#tls-alpn-01).
+:::caution Autocert usage limits
```
kenjenkins, line 46: This part is really specific to Let's Encrypt, not to Autocert in general.

kenjenkins (commented):
Hey Zach, thanks for taking this on. I've got a lot of thoughts about this. Let me know if you'd prefer to go over this one together.

content/docs/reference/autocert.mdx
```diff
@@ -2,10 +2,12 @@
 id: autocert
 title: Autocert Settings
 description: |
-  Enable Autocert and Autocert settings to retrieve, manage, and renew public facing TLS certificates from Let's Encrypt.
+  Enable Autocert and Autocert settings to retrieve, manage, and renew public-facing TLS certificates from Let's Encrypt.
```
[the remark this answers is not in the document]
ZPain8464: I forgot to update the front matter here. I'll commit your suggestion for now but may tweak it after further review.

content/docs/reference/autocert.mdx
ZPain8464: My reasoning originally was that the Let's Encrypt article links to the RFC, so the user would likely follow that path if they really wanted to  know. But, I didn't consider the fact that you can configure your preferred CA. Linking to the source of truth and ultimate authority here makes more sense.

content/docs/reference/autocert.mdx
ZPain8464: The original definition included "the authenticate service".  Wouldn't users want to know that it's not supported when using the hosted authn service?

content/docs/reference/autocert.mdx
ZPain8464: I feel like there are two approaches to this:
- Include this information at the top and less details in the settings themselves. This is ideal assuming most users start at the top and work their way down.
- Include the pertinent details under the settings themselves, not the top, in case users enter a page by Algolia search or a reference page filtered search. In this case, they'd miss the details at the top.

I'd imagine most users would land on the page and scroll to the setting without reading the summary, or be directed to a setting from the reference page or Algolia search  to read how to configure it. 

In the case of Autocert and Autocert Directory, they'd then see that Autocert must be used in conjunction with the Autocert Directory setting. If you provide this at the top of the page, users might skim over it or miss it entirely. 

I'm not opposed to moving it to the top, but this is something I considered when making changes. What are your thoughts here?

ZPain8464: The more I think about it, though, logically you'd expect to see the most critical details at the top of a page, not buried under a setting you may miss. Happy to discuss more on Slack or a quick call, though.

content/docs/reference/autocert.mdx
ZPain8464: Ah OK, I wasn't sure if the "internet accessible" detail was important here, or if it just meant an available, open port. Seems like they're saying the same thing. I appreciate the clarification. I'll update.

content/docs/reference/autocert.mdx
ZPain8464: Good point. I think since we use LE by default, it still makes sense to include this information somewhere. Maybe we could change the admonition title to "Let's Encrypt usage limits" and change the text say something along the lines of, "If you use the default Let's Encrypt CA with Autocert, you agree to LE's subscriber agreement."

content/docs/reference/autocert.mdx
ZPain8464: I updated this section to remove the system trust store detail.

content/docs/reference/autocert.mdx
ZPain8464: Updated.

kenjenkins (approved):

PR #1532, opened 2024-07-22
Updates configuration page

ZPain8464:
This PR update the location and some of the content in the "Configuration" page. This page was originally in `/docs/internals/configuration`. 

I've intentionally left the "All In One and Split Service modes" sections sparse, as this a Draft PR. We could expand on this section in a separate capabilities page or even a guide, but for now, I didn't change much besides adding a "Our recommendation" tip, which could also be fine tuned. 

TODOs:
- Fix breaking anchors

Resolves https://github.com/pomerium/internal/issues/1853


content/docs/core/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Configuration syntax
+
+Both configuration file keys and environment variables are case sensitive.
+
+Configuration file keys are always lowercase and use dashes (`-`). Environmental variables are identical to configuration file keys, except they are always uppercase and use underscores (`_`).
+
+See the [Reference](/docs/reference) page for a comprehensive list of Pomerium's configuration settings.
+
+## All In One and Split Service modes
```
kenjenkins, line 32: Actually, should we have a separate page to talk about all-in-one and split service mode? This almost feels like a different topic to me.

@desimone would love to hear your thoughts on this.

kenjenkins (commented):
Thanks, I think this is a step in the right direction. I've left some comments, but please let me know if you're looking for more high-level feedback first.

content/docs/core/configuration.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Configuration
+description: This page discusses configuration settings for Pomerium Core.
+keywords: [core configuration]
+sidebar_label: Configuration
+---
+
+# Configuration
+
+You can configure Pomerium using either a configuration file or [environmental variables](https://en.wikipedia.org/wiki/Environment_variable). If using a configuration file, the following languages are supported:
```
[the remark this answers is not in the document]
ZPain8464: Agree. Thanks for calling that out.

content/docs/core/configuration.md
```diff
@@ -0,0 +1,57 @@
...
+
+- [YAML](https://yaml.org/)
+- [JSON](https://www.json.org/json-en.html)
+- [TOML](https://toml.io/en/)
```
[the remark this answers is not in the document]
ZPain8464: Good point, I'll update.

content/docs/core/configuration.md
ZPain8464: Yeah, if we moved this topic to a separate page, we could offload some of the content in [High Availability](https://www.pomerium.com/docs/capabilities/high-availability) as well.

It begs the question of where we'd put a All In One vs Split Services mode page, though. Capabilities doesn't seem like the best place.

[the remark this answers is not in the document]
ZPain8464: So this is a style question, and a product one. We're very inconsistent in our docs with either of these modes. It's frequently used as a compound adjective, in which case "all-in-one mode" or "split-service mode" makes sense. 

Take [this page](https://www.pomerium.com/docs/core/upgrading#new-cache-service) for example:
"- For all-in-one deployments, no changes are required...
- For split-service deployments, you will need to deploy an additional service called cache..."

This is fine in this context. 

When we refer to is as a deployment type, it's actually a noun, if not a proper noun. For example, [this page](https://www.pomerium.com/docs/reference/service-urls#authenticate-internal-service-url) shows the inconsistency and how it's used as a noun:
"The other service URL settings are needed only for split service deployments. When running in all-in-one mode (which is the recommended mode), these settings are not needed."

We have the liberty to decide how it should be used in our docs. In my opinion, these modes should be referred to as proper nouns, so "All In One mode" or "Split Service mode". I don't mind changing it to "All-in-One" mode and "Split-Service" mode, or "All in One" mode (we don't have to capitalize "in" either) and "Split Service" mode, but we should define it and stick to it.

I can be lazy about it as well. This is where a style guide would help enforce the decision. Thoughts?

desimone: > should we have a separate page to talk about all-in-one and split service mode? This almost feels like a different topic to me.

Yes, I think that makes a lot of sense.

ZPain8464: I've created a separate issue for work on a new All in One vs Split Services Mode page:
- https://github.com/pomerium/internal/issues/1879

ZPain8464:
Hey @kenjenkins, I created a ticket for a new All in one vs. Split services mode page, but I moved it to backlog for now. To move this along, I also removed the dashes you mentioned earlier. We can address style guide decisions in the other ticket.

content/docs/capabilities/high-availability.mdx
```diff
@@ -10,7 +10,7 @@ import InstallMkcert from '@site/content/_install-mkcert.md';
 
 Pomerium is designed to be run in two modes: All-In-One or Split Service. These modes are not mutually exclusive, meaning you can run one or multiple instances of Pomerium in all-in-one mode, and spin up additional instances for specific components as needed.
 
-Each instance of Pomerium runs in all-in-one mode unless specified to run as a specific component by the `services` key. See [All-In-One vs. Split Service mode](/docs/internals/configuration#all-in-one-vs-split-service-mode) for more details.
+Each instance of Pomerium runs in all-in-one mode unless specified to run as a specific component by the `services` key. See [All-In-One vs. Split Service mode](/docs/core/configuration#all-in-one-and-split-service-modes-split-service-mode) for more details.
```
kenjenkins, line 13: Does this link target exist? Should it be `#all-in-one-and-split-service-modes` instead?

(ditto throughout this PR)

content/docs/core/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+
+Pomerium can hot-reload route configuration details, authorization policy, certificates, and other proxy settings.
+
+:::
+
+### Configuration syntax
+
+Both configuration file keys and environment variables are case sensitive.
+
+Configuration file keys are always lowercase. Environmental variables are identical to configuration file keys, except they are always uppercase.
```
kenjenkins, line 30: Environmental -> environment here if we want to be consistent with this.

content/docs/core/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+
+See the [Reference](/docs/reference) page for a comprehensive list of Pomerium's configuration settings.
+
+## All-In-One and Split-Service modes
+
+You can configure Pomerium using either All-In-One mode or Split Service mode.
+
+### All-In-One mode
+
+All-In-One mode means all of Pomerium's configuration settings are set in a single configuration file. You can use All-In-One mode when running Pomerium:
```
kenjenkins, line 40: For me this is not a clear explanation of the difference between all-in-one and split service mode.

In all-in-one mode, a single Pomerium process runs all of the four logical services. In split service mode, each logical service runs as its own process.

kenjenkins, line 40: I would prefer we pick one way to write "all-in-one" mode and use it consistently. By my count the existing docs have:
- 49 instances of "all-in-one"
- 3 instances of "All-in-One"
- 1 instance each of "All-in-one" and "All-in-One"

Based on the existing usage "All-In-One" looks out of place to me here.

content/docs/capabilities/high-availability.mdx
ZPain8464: Updated.

content/docs/core/configuration.md
ZPain8464: I missed this instance and updated it for this page. But, there are many instances elsewhere in our docs that use "environmental". I'll note this decision for future PRs, though.

content/docs/core/configuration.md
ZPain8464: Updated.

ZPain8464: OK. I left a comment earlier which explains my position on this. But, I've updated this page so that is uses "All-in-One".

content/docs/core/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+## All-in-One and Split-Service modes
+
+You can configure Pomerium using either All-in-One mode or Split-Service mode.
+
+### All-in-One mode
+
+All-in-One mode means a single Pomerium process runs all of the [four logical services](/docs/internals/architecture#component-level). You can use All-in-One mode when running Pomerium:
+
+- As a single system service or container, or
+- In a distributed environment where there are multiple processes that each handle separate [Pomerium services](/docs/internals/architecture#component-level).
```
kenjenkins, line 43: I don't understand what this is saying. In all-in-one mode the services are combined, not separate.

content/docs/core/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+
+All-in-One mode is the default configuration mode, and the easiest way to configure Pomerium.
+
+### Split-Service mode
+
+Alternately, you can create individual configuration files (or sets of environment variables) for each Pomerium service. In Split-Service mode, each configuration file (or set of environment variables) defines which service a process will run by using the [service mode](/docs/reference/service-mode) key.
+
+:::tip Our recommendation
+
+We recommend All-in-One mode to configure Pomerium for the following reasons:
```
kenjenkins, line 53: The word order here feels backwards to me.

```suggestion
We recommend running Pomerium in all-in-one mode for the following reasons:
```

content/docs/core/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+All-in-One mode means a single Pomerium process runs all of the [four logical services](/docs/internals/architecture#component-level). You can use All-in-One mode when running Pomerium:
+
+- As a single system service or container, or
+- In a distributed environment where there are multiple processes that each handle separate [Pomerium services](/docs/internals/architecture#component-level).
+
+All-in-One mode is the default configuration mode, and the easiest way to configure Pomerium.
+
+### Split-Service mode
+
+Alternately, you can create individual configuration files (or sets of environment variables) for each Pomerium service. In Split-Service mode, each configuration file (or set of environment variables) defines which service a process will run by using the [service mode](/docs/reference/service-mode) key.
```
kenjenkins, line 49: I think it would be good to start with a brief explanation of how split service mode differs from all-in-one mode. Probably also good to link to the Service URL Settings page here. Maybe something like this?

```suggestion
Alternatively, you can run each Pomerium service as its own process. You might want to do this in order to scale each service independently, or so that you can enforce different permissions on each service.

To configure Pomerium in this mode, you will need to create individual configuration files (or sets of environment variables) for each Pomerium service. Each configuration should set the [Service Mode](/docs/reference/service-mode) option to tell Pomerium which service to run. You will also need the [Service URL Settings](/docs/reference/service-urls) to tell each Pomerium service how to communicate with the other Pomerium services.
```

content/docs/core/configuration.md
kenjenkins: For the record, I'm saying "all-in-one" (lowercase) would be most consistent with the existing documentation. If you don't think we should aim to be consistent I won't argue the point any further.

PR #1533, opened 2024-07-24
Adds envoy-error-messages header id

ZPain8464:
This PR adds an [explicit heading ID](https://docusaurus.io/docs/next/markdown-features/toc#heading-ids) to the [Upstream connection errors](https://www.pomerium.com/docs/troubleshooting#upstream-connection-errors) section on the Troubleshooting page. This should fix the linking issue described in https://github.com/pomerium/pomerium/issues/5189.

Resolves https://github.com/pomerium/pomerium/issues/5189


kenjenkins (approved):
Sounds good, thanks!

PR #1535, opened 2024-07-25
Updates secrets settings

ZPain8464:
This PR consolidates the Shared Secret/File and Signing Key/File settings into two reference pages:

- Shared Secret Settings
- Signing Key Settings

Resolves https://github.com/pomerium/internal/issues/1856


content/docs/reference/shared-secret.mdx
```diff
@@ -1,8 +1,7 @@
 ---
-id: shared-secret
-title: Shared Secret
-description: |
-  Shared Secret is the base64 encoded 256-bit key used to mutually authenticate requests between services.
+id: shared-secret-settings
+title: Shared Secret Settings
```
kenjenkins, line 3: If it were up to me I would keep the existing title and URL path here. This is really just two ways to configure the same setting.

```suggestion
id: shared-secret
title: Shared Secret
```

And likewise for signing key.

content/docs/reference/shared-secret.mdx
```diff
@@ -14,33 +13,39 @@ toc_max_heading_level: 2
...
+# Shared Secret Settings
 
-## Summary
+## Shared Secret
 
-**Shared Secret** is the base64-encoded, 256-bit key used to mutually authenticate requests between services. It's critical that secret keys are random, and stored safely.
+**Shared Secret** is the base64-encoded, 256-bit key used to mutually authenticate requests between Pomerium services. It's critical that secret keys are random, and stored safely.
 
-## How to configure
+### How to configure
```
kenjenkins, line 22: I think we should follow https://www.pomerium.com/docs/reference/certificates#certificate-authority as an example here. That is, put `shared_secret` and `shared_secret_file` together as two rows in the same "How to configure" table.

Ditto for signing key.

content/docs/reference/shared-secret.mdx
```diff
@@ -14,33 +13,39 @@ toc_max_heading_level: 2
...
 <Tabs>
 <TabItem value="Core" label="Core">
 
 | **Config file keys** | **Environment variables** | **Type** | **Usage** |
 | :-- | :-- | :-- | :-- |
-| `shared_secret` | `SHARED_SECRET` | `string` | **required** (unless using [shared_secret_file](/docs/reference/shared-secret-file)) |
+| `shared_secret` | `SHARED_SECRET` | `string` | \***optional** |
 
-:::tip **Note**
+\* Standalone Pomerium Core configurations do not require a `shared_secret` or `shared_secret_file`. If you don't set a shared secret, Pomerium will generate one for you.
```
kenjenkins, line 31: I don't know if this is completely accurate. I believe you would also need to set the same shared secret on each service if running in split service mode.

(Feel free to leave this for a separate follow-up item.)

content/docs/reference/shared-secret.mdx
```diff
@@ -14,33 +13,39 @@ toc_max_heading_level: 2
...
 
 :::
 
 ### Examples
 
+:::note
+
+If you adjust your shared secret and/or how it's accessed by Pomerium, you may create a [**secret mismatch**](/docs/troubleshooting#shared-secret-mismatch).
+
+:::
```
kenjenkins, line 47: Seems odd to put this as the first thing below "Examples." Maybe move it up to the summary section?

kenjenkins (commented):
I think we can simplify this even further.

content/docs/capabilities/getting-users-identity.md
```diff
@@ -167,5 +167,5 @@ In an actual client, you'll want to ensure that all the other claims values are
 [key management service]: https://en.wikipedia.org/wiki/Key_management
 [nist p-256]: https://csrc.nist.gov/csrc/media/events/workshop-on-elliptic-curve-cryptography-standards/documents/papers/session6-adalier-mehmet.pdf
 [pass identity headers]: /docs/reference/routes/pass-identity-headers-per-route
-[signing key]: /docs/reference/signing-key
-[signing key file]: /docs/reference/signing-key-file
+[signing key]: /docs/reference/signing-key#signing-key
+[signing key file]: /docs/reference/signing-key#signing-key-file
```
kenjenkins, line 171: I think these can both link to the same place now that these are combined into one.
```suggestion
[signing key]: /docs/reference/signing-key
[signing key file]: /docs/reference/signing-key
```

Or better yet, maybe line 101 can be simplified to just

```
In order to use the `/.well-known/pomerium/jwks.json` endpoint you must set the [Signing Key] configuration option.
```

and then we can delete line 171 here.

content/docs/core/quickstart.md
```diff
@@ -71,7 +71,7 @@ Although identity verification failed, you successfully integrated Pomerium with
 
 :::tip
 
-Because this guide doesn't include a [signing key](/docs/reference/signing-key) in the configuration, identity verification will fail.
+Because this guide doesn't include a [signing key](/docs/reference/signing-key#signing-key) in the configuration, identity verification will fail.
```
kenjenkins, line 74: I think these anchor tag changes can be reverted too, here and in the following files too?

kenjenkins (commented):
Thanks! Just a few more comments.

content/docs/capabilities/getting-users-identity.md
ZPain8464: Done.

content/docs/core/quickstart.md
ZPain8464: I searched for anchors with `#signing-key` and removed it. Good call.

content/docs/reference/shared-secret.mdx
ZPain8464: OK. I created an issue so we can track work on this: #1538

content/docs/capabilities/getting-users-identity.md
```diff
@@ -98,7 +98,7 @@ curl https://your-app.corp.example.com/.well-known/pomerium/jwks.json | jq
 
 :::caution
 
-In order to use the `/.well-known/pomerium/jwks.json` endpoint you must set either the [Signing Key] or [Signing Key File] configuration option.
+In order to use the `/.well-known/pomerium/jwks.json` endpoint, you must set either [Signing Key] configuration option.
```
kenjenkins, line 101: nit: I think conceptually we can think of "Signing Key" as a single option

```suggestion
In order to use the `/.well-known/pomerium/jwks.json` endpoint, you must set the [Signing Key] configuration option.
```

content/docs/internals/data-storage.md
```diff
@@ -94,4 +94,4 @@ Configuration options for each backend are detailed in the [**databroker configu
 
 ## Troubleshooting
 
-Most issues with the Databroker service are caused by a [`shared_secret`](/docs/reference/shared-secret) mismatch between services. See [Troubleshooting - Shared Secret Mismatch](/docs/reference/shared-secret) for details.
+Most issues with the Databroker service are caused by a [`shared_secret`](/docs/reference/shared-secret#shared-secret) mismatch between services. See [Troubleshooting - Shared Secret Mismatch](/docs/reference/shared-secret#shared-secret) for details.
```
kenjenkins, line 97: We should revert this change now as well, right?

(I'll go ahead and push a commit to do that.)

content/docs/troubleshooting.mdx
```diff
@@ -148,7 +148,7 @@ Events:
 
 ### Shared Secret Mismatch
 
-Pomerium's independent services communicate securely using a [shared secret](/docs/reference/shared-secret). When services or the databroker have mismatched secrets, Pomerium will fail.
+Pomerium's independent services communicate securely using a [shared secret](/docs/reference/shared-secret#shared-secret). When Pomerium services share a mismatched secret, or these services share a secret that is not the same secret set in the Databroker service, Pomerium will fail.
```
kenjenkins, line 151: I'm not sure this is an improvement. "Share" feels like the wrong verb to use if we're talking about different secrets set on different services.

kenjenkins (approved):

PR #1543, opened 2024-07-31
Adds Rego capabilities page

ZPain8464:
This PR: 
- Adds a new Custom Rego Policies capabilities page.
- Alphabetizes the ENT capabilities
- Moves Rego support content to the new Rego page

Resolves https://github.com/pomerium/internal/issues/1874


content/docs/capabilities/rego.mdx
```diff
@@ -0,0 +1,169 @@
...
+} else = [false, {"custom-rego-unauthorized"}] {
+  session.id != ""
+} else = [false, {"user-unauthenticated"}] {
+  true
+}
+```
+
+This example pulls session data from the Databroker service using `type.googleapis.com/session.Session` for users and `type.googleapis.com/user.ServiceAccount` for service accounts.
+
+::::enterprise
```
kenjenkins, line 157: I think we don't need a second Enterprise admonition here if there's already one at the top of the page. That would also allow us to avoid nesting a "Note" admonition inside another admonition.

What about moving this information up, before the "Outputs" section heading?

kenjenkins (approved):
Looks good, just one comment.

content/docs/capabilities/rego.mdx
ZPain8464: Agreed. I went ahead and moved this section up, and removed the enterprise admonition wrapper. Let me know if that works.

PR #1550, opened 2024-08-08
Fixes typo in mTLS Clients guide

ZPain8464:
On [this page](https://www.pomerium.com/docs/capabilities/mtls-clients#install-your-client-certificate), we tell the user to import the `_wildcard` cert created from earlier in the guide. They should be importing the client cert instead. This PR updates the text to mention the correct cert.

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -87,6 +87,7 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 /docs/guides/securing-tcp /docs/capabilities/tcp
 /docs/guides/tcp.html /docs/guides/securing-tcp
 /guides/vs-code-server.html /docs/guides/code-server
+/docs/guides/vs-code-server.html /docs/guides/code-server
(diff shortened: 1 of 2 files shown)


static/_redirects
```diff
@@ -87,6 +87,7 @@ https://0-20-0.docs.pomerium.com/category/guides https://0-20-0.docs.pomerium.co
 /docs/guides/securing-tcp /docs/capabilities/tcp
 /docs/guides/tcp.html /docs/guides/securing-tcp
 /guides/vs-code-server.html /docs/guides/code-server
+/docs/guides/vs-code-server.html /docs/guides/code-server
```
kenjenkins, line 90: This seems unrelated? I think this was from another PR that was merged already, so it should disappear if you sync the branch.

kenjenkins (approved):
Looks good!

static/_redirects
ZPain8464: Ah definitely a mistake. Thanks for pointing that one out.

PR #1552, opened 2024-08-09
Adds jti recommendation

ZPain8464:
This PR adds a tip admonition explaining how the `jti` claim can be used to prevent session replay attempts. 

Page: https://pomerium.com/docs/capabilities/getting-users-identity

Related to https://github.com/pomerium/internal/issues/1876

### DIFF
--- a/content/docs/capabilities/getting-users-identity.md
+++ b/content/docs/capabilities/getting-users-identity.md
@@ -220,6 +220,12 @@ The Pomerium JWT contains at least the following claims:
 | `groups` | The user's group memberships (if supported for the identity provider). |
 | `name` | The user's full name, as specified by the identity provider. |
 
+:::tip Prevent session replay attempts
+
+The `jti` claim (the JWT ID) contains a unique identifier assigned to each Pomerium JWT. You can prevent session replay attempts by implementing a system that monitors `jti` reuse either in real time or by checking session logs.
+
+:::
+
 ### JWT Settings
 
 Use these settings to configure Pomerium to forward the Pomerium JWT to upstream services:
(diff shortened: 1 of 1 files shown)


content/docs/capabilities/getting-users-identity.md
```diff
@@ -220,6 +220,12 @@ The Pomerium JWT contains at least the following claims:
 | `groups` | The user's group memberships (if supported for the identity provider). |
 | `name` | The user's full name, as specified by the identity provider. |
 
+:::tip Prevent session replay attempts
+
+The `jti` claim (the JWT ID) contains a unique identifier assigned to each Pomerium JWT. You can prevent session replay attempts by implementing a system that monitors `jti` reuse either in real time or by checking session logs.
```
ZPain8464, line 225: This might come across as a little prescriptive.  Wondering if we should change it to "One way you can prevent session replay attempts is by..."

kenjenkins: Might make sense to distinguish two cases: prevent vs. detect. If you can check the `jti` value in real time, you may be able to use it to _prevent_ a replay attempt. _Or_, if you log the `jti` value, you can at least _detect_ replayed JWTs after the fact.

kenjenkins, line 225: One other comment — the change to unique `jti` values has been merged but not yet released, so I think this should not be backported to the v0.26 docs site.

We should probably also correct the information in the `jti` row of the table above — in v0.26 and earlier this contains the Pomerium session ID.

ZPain8464: I'll work with that wording. Thanks for the suggestion.

ZPain8464: OK. I'll keep this in mind going forward. I can add the previous session ID description and keep these changes in `main`.

kenjenkins (approved):

PR #1570, opened 2024-08-22
updates vNext versions dropdown label

ZPain8464:
Updates the vNext label in the Docusaurus config file so that it says "upcoming release".  I also updated the latest release label to say "latest release".

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -57,7 +57,7 @@ const config = {
           },
           versions: {
             current: {
-              label: 'vNext (current)',
+              label: 'vNext (upcoming release)',
               badge: true,
             },
           },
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -164,7 +164,7 @@ const config = {
           dropdownItemsAfter: [
             {
               to: 'https://0-26-0.docs.pomerium.com/docs',
-              label: 'v0.26 (latest)',
+              label: 'v0.26 (latest release)',
```
kenjenkins, line 167: Personal preference: I would keep this as just "(latest)" to avoid making the dropdown quite so wide.

kenjenkins (approved):

docusaurus.config.js
ZPain8464: Thanks for the feedback. I agree here. I'll remove "latest".

PR #1577, opened 2024-09-03
update runtime flags

kenjenkins:
Alphabetize existing flags, update the `envoy_resource_manager` flag name, and document the new `config_hot_reload `flag. (New in v0.27, so this should not be backported.)

Related:
- https://github.com/pomerium/pomerium/pull/5255
- https://github.com/pomerium/pomerium/issues/5079


content/docs/reference/runtime-flags.md
```diff
@@ -34,10 +34,11 @@ The available flags are:
 
 | Runtime Flag | Description | Default |
 | :-- | :-- | :-- |
+| `config_hot_reload` | Enables automatic config reloading triggered whenever a configuration file is written to (either the main Pomerium configuration file or a file referenced from the main configuration). In some rare cases this may not work correctly, so this option provides a way to disable this behavior. (See issue [#5079](https://github.com/pomerium/pomerium/issues/5079) for more context.) | `true` |
```
ZPain8464, line 37:
```suggestion
| `config_hot_reload` | Enables automatic config reloading triggered whenever a configuration file is written to (either the main Pomerium configuration file or a file referenced from the main configuration). In some rare cases this may not work correctly, so this setting provides a way to disable this behavior. (See issue [#5079](https://github.com/pomerium/pomerium/issues/5079) for more context.) | `true` |
```

ZPain8464: Small nit: Can we change "option" to "setting" so it's consistent with the rest of this page? Of course if there's a good reason to keep this as "option", then we don't need to update. 

I'll go ahead and approve though. Thanks for documenting this.

ZPain8464 (approved):

content/docs/reference/runtime-flags.md
kenjenkins: Sure, done.

PR #1583, opened 2024-09-09
document databroker_storage_connection_string_file

kenjenkins:
Starting in v0.27, the databroker storage connection string may be read from a file rather than provided directly in the configuration (e.g. in case you want to store this data in a [Docker Secret](https://docs.docker.com/engine/swarm/secrets/) or a [Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/).)

Related issue:
- https://github.com/pomerium/pomerium/issues/5228


ZPain8464 (approved):
Thanks for documenting this, @kenjenkins. The copy and examples look good to me.

PR #1584, opened 2024-09-09
Removes /.pomerium/jwt endpoint instances

ZPain8464:
This PR removes the `/.pomerium/jwt` instances from our documentation.

I also made a decision to remove the `PomeriumVerifier` reference from https://www.pomerium.com/docs/guides/verify-jwt#pomeriumverifier-reference since there were several changes to the js-sdk repo.


ZPain8464:
@kenjenkins there is still a mention of the JWT endpoint in our Advanced Policies guide, but I plan to update that after we merge the new runtime flag.

content/docs/guides/verify-jwt.md
```diff
@@ -62,32 +62,6 @@ The `issuer` and `audience` parameters should both be set to the domain of the *
   </div>
 </details>
 
-### PomeriumVerifier reference
```
kenjenkins, line 65: I think we should either remove the entire single-page web app section, or else switch it to use the server-side example from the js-sdk repo: https://github.com/pomerium/js-sdk/blob/main/examples/express/index.js.

ZPain8464: The easiest lift right now would probably be to remove that section altogether. Why swap the front-end example for a server-side example?

kenjenkins: Sure, sounds good to me.

Eventually I think it could make good sense to have both server-side Go and JS examples as those are the two SDKs we have right now.

kenjenkins (approved):

PR #1587, opened 2024-09-09
document new h2c:// scheme

kenjenkins:
Starting in v0.27 the route `to` URL supports a new `h2c://` scheme, for HTTP/2 cleartext upstream services.

Related issue:
- https://github.com/pomerium/pomerium/issues/5213

### DIFF
--- a/content/docs/reference/routes/to.mdx
+++ b/content/docs/reference/routes/to.mdx
@@ -23,9 +23,9 @@ import TabItem from '@theme/TabItem';
 <Tabs>
 <TabItem value="Core" label="Core">
 
-| **YAML**/**JSON** setting | **Type** | **Usage**    | **Schemes**                   |
-| :------------------------ | :------- | :----------- | :---------------------------- |
-| `to`                      | `URL`    | **optional** | `http`, `https`, `h2c`, `tcp` |
+| **YAML**/**JSON** setting | **Type** | **Usage** | **Schemes** |
+| :-- | :-- | :-- | :-- |
+| `to` | `URL` | **optional** | `http`, `https`, `h2c`, `tcp` |
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/routes/to.mdx
```diff
@@ -81,6 +81,20 @@ A load balancing weight may be associated with a particular upstream by appendin
   to: ['http://a,10', 'http://b,20']
 ```
 
+### HTTP/2 cleartext
+
+When Pomerium connects to an `https` upstream, it will negotiate either HTTP/1.1 or HTTP/2 using [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation) (as part of the TLS handshake).
+
+To configure Pomerium to make requests to an upstream service using HTTP/2 _without_ TLS (i.e. in cleartext), use the special `h2c://` scheme.
```
ZPain8464, line 88: Small nit: I try to avoid latinate expressions entirely, such as "i.e." (this is covered in a lot of style guides, for example [Google's developer docs style guide](https://developers.google.com/style/abbreviations#dont-use)). Let's update to:
```suggestion
To configure Pomerium to make requests to an upstream service using HTTP/2 _without_ TLS (that is, in cleartext), use the special `h2c://` scheme.
```

content/docs/reference/routes/to.mdx
```diff
@@ -81,6 +81,20 @@ A load balancing weight may be associated with a particular upstream by appendin
...
+
```
ZPain8464, line 89: Should we add an example? Pulling from Joe's PR:
```suggestion
routes:
  - from: https://grpc.localhost.pomerium.io
    to: h2c://localhost:9090
```

ZPain8464 (approved):
@kenjenkins I left two suggestions in this PR. Thanks for documenting it.

content/docs/reference/routes/to.mdx
ZPain8464: If we do include an example, we should place it above the "info" admonition.

content/docs/reference/routes/to.mdx
kenjenkins: Sure, sounds good.

content/docs/reference/routes/to.mdx
kenjenkins: Sounds good; changed the `from` address to https://example.com for consistency with the other examples in the preceding sections.

PR #1592, opened 2024-09-11
Adds v0.27 Pomerium Core upgrade guide

ZPain8464:
This PR: 
- Adds an upgrade guide for Core with callouts about the `/.pomerium/jwt` deprecation and upgrading requirement changes for Pomerium Zero deployments in Kubernetes.
- Adds a note admonition on the [Zero Install - Upgrade Pomerium Zero](https://www.pomerium.com/docs/zero/install#upgrade-pomerium-zero) page in the Kubernetes tab that instructs users to remove their StatefulSet before upgrading.  Links to the upgrade guide. (This note needs to be removed in v0.28.)

@wasaga I know we discussed the Kubernetes upgrade steps together, but tagging you for review just in case I missed anything here.


content/docs/zero/install.mdx
```diff
@@ -93,6 +60,39 @@ docker compose up -d
...
+
+Pomerium requires an internet-accessible server and the ability to bind port 443.
+
+</TabItem>
+<TabItem value="kubernetes" label="Kubernetes">
+
+First, install Pomerium into your Kubernetes cluster:
+
+```bash
+kubectl apply -k github.com/pomerium/pomerium/k8s/zero\?ref=v0.26.0
```
kenjenkins, line 80: @wasaga does this need to be `ref=main` for now, to match https://github.com/pomerium/pomerium-zero/pull/2972?

content/docs/zero/install.mdx
```diff
@@ -18,39 +18,6 @@ If you haven't signed up for Pomerium Zero, you can [**create an account**](http
 :::
 
 <Tabs>
-<TabItem value="linux" label="Linux">
```
kenjenkins, line 21: The order of these tabs in the onboarding flow is still Linux, Kubernetes, Docker, Manual. Would it make sense to keep the same order in the docs?

kenjenkins (approved):

content/docs/zero/install.mdx
ZPain8464: That makes sense. I'll update to so it's consistent with the onboarding flow.

content/docs/zero/install.mdx
ZPain8464: I'll be updating version tags throughout the docs as part of the docs release process. I've added this one to the list.

PR #1594, opened 2024-09-11
Documents validation_mode setting Enterprise setting

ZPain8464:
This PR: 

- Adds the new `validation_mode` configuration option to the [Enterprise Configuration](https://www.pomerium.com/docs/enterprise/configure) page. 
- Adds a "(deprecated)" parenthetical in the `disable_validation` description table.

TODO: After merging, add a link from the "(deprecated)" parenthetical note to the Console changelog.

### DIFF
--- a/content/docs/enterprise/configure.mdx
+++ b/content/docs/enterprise/configure.mdx
@@ -67,6 +67,7 @@ All values are case sensitive unless otherwise noted.
 | <a className="entRef-anchor" id="tls-key">#</a><a href='#tls-key'>TLS_KEY</a> | base64-encoded string of tls-key | none |
 | <a className="entRef-anchor" id="tls-key-file">#</a><a href='#tls-key-file'>TLS_KEY_FILE</a> | file storing tls-key | none |
 | <a className="entRef-anchor" id="use-static-assets">#</a><a href='#use-static-assets'>USE_STATIC_ASSETS</a> | When false, forward static requests to `localhost:3000`. | `true` |
+| <a className="entRef-anchor" id="validation_mode">#</a><a href='#validation_mode'>VALIDATION_MODE</a> | Validates config based on the specified mode: `full` (The default config validation mode), `static` (A "light" validation mode that catches most config issues), or `none` (Disables config validation).  | `full` |
 
 </TabItem>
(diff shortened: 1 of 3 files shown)


content/docs/enterprise/configure.mdx
```diff
@@ -43,7 +43,7 @@ All values are case sensitive unless otherwise noted.
 | <a className="entRef-anchor" id="databroker-service-url">#</a><a href='#databroker-service-url'>DATABROKER_SERVICE_URL</a> | The databroker service URL. | `http://localhost:5443` |
 | <a className="entRef-anchor" id="debug-config-dump">#</a><a href='#debug-config-dump'>DEBUG_CONFIG_DUMP</a> | Dumps the Databroker configuration. This is a debug option to be used only when specified by Pomerium Support. | `false` |
 | <a className="entRef-anchor" id="disable-remote-diagnostics">#</a><a href='#disable-remote-diagnostics'>DISABLE_REMOTE_DIAGNOSTICS</a> | Disable remote diagnostics. | `true` |
-| <a className="entRef-anchor" id="disable-validation">#</a><a href='#disable-validation'>DISABLE_VALIDATION</a> | Disable config validation. | `false` |
+| <a className="entRef-anchor" id="disable-validation">#</a><a href='#disable-validation'>DISABLE_VALIDATION</a> | (deprecated) Disable config validation. | `false` |
```
kenjenkins, line 46: I think it would be nice to mention the replacement here, maybe something like this?

```suggestion
| <a className="entRef-anchor" id="disable-validation">#</a><a href='#disable-validation'>DISABLE_VALIDATION</a> | (deprecated, please update your configuration to set `VALIDATION_MODE=none` instead) Disable config validation.  | `false` |
```

(and similarly, below)

kenjenkins (approved):

content/docs/enterprise/configure.mdx
ZPain8464: I originally tried hyperlinking to the new setting in the parenthetical, but it was a bit busy, like:
`(deprecated; see validation_mode)`.

I do think it needs to direct users to the fix, though, and it will look "busy" either way.  Your suggestion also tells them how to achieve the same effect with the new option, so let's add it.

PR #1611, opened 2024-09-19
Adds a pushfeedback widget

ZPain8464:
This PR adds the pushfeedback widget to our docs pages. I added the `hideEmail` property to remove the email input, but we can also include the input if we want to support that option.

Tagging @gaurdro here in case he has any input as well.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -273,6 +273,15 @@ const config = {
       },
     ],
     './docusaurus-plugins/src/webpackDebugFix.js',
+    [
+      'docusaurus-pushfeedback',{
+          project: 'esgee0kxds',
+          buttonPosition: 'center-right',
+          modalPosition: 'sidebar-right',
+          buttonStyle: 'dark',
+          modalTitle: 'Share your thoughts'
+      }
+    ]
   ],
 };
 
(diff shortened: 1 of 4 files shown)


docusaurus.config.js
```diff
@@ -273,6 +273,16 @@ const config = {
...
     ],
     './docusaurus-plugins/src/webpackDebugFix.js',
+    [
+      'docusaurus-pushfeedback',{
+          project: 'esgee0kxds',
+          buttonPosition: 'center-right',
+          modalPosition: 'sidebar-right',
+          buttonStyle: 'dark',
+          modalTitle: 'Share your thoughts',
+          hideEmail: true,
```
rjbeers, line 283: whether we hide or not is not worth blocking this PR, but in my opinion, having the ability to optionally provide an email address would be good to try now or later to see if we can collect emails that we haven't seen before from open source users.

gaurdro (approved):
LGTM, either way.    If we request an email it would fall under visitors (marketing) in our [privacy policy](https://www.pomerium.com/privacy-policy), which I think is acceptable given the listed reason why. 

> To share information about our products and/or services.

ZPain8464:
I'll update to include the email input. Thanks for taking a look.

PR #1634, opened 2024-10-03
remove gRPC Client DNS RoundRobin setting

kenjenkins:
Related issue:
- https://github.com/pomerium/pomerium/issues/5307


ZPain8464:
We should also remove the entry from the `reference.json` file:

```
  "grpc-client-dns-roundrobin": {
    "id": "grpc-client-dns-roundrobin",
    "title": "gRPC Client DNS RoundRobin",
    "path": "/grpc#grpc-client-dns-roundrobin",
    "services": [],
    "type": "bool"
  },
```

ZPain8464 (approved):
Thanks for removing this.

Issue #1299, opened 2024-03-07
Add a note on the path rewriting page to clarify that Pomerium does not modify upstream responses


kenjenkins:
Page: https://www.pomerium.com/docs/reference/routes/path-rewriting

## What's incorrect or missing

We've had a couple recent issue reports that path rewriting does not behave as users expect:
 - https://github.com/pomerium/pomerium/issues/5020
 - https://github.com/pomerium/pomerium/issues/4909

As well as a similar Slack thread last month:
 - https://pomerium-io.slack.com/archives/CK92MUAES/p1707343997402799

## What's the resolution?

I think it would be good to add some note or callout on this page to mention that just rewriting a path prefix in Pomerium is often not enough to serve a particular upstream service under a path prefix. Many upstream services will also need to be configured appropriately, so that they add this path prefix to any link URLs, HTTP redirects, etc.

We could even add a few examples of relevant configuration options, to help users get a sense of what to look for when they search the documentation for the service they are using. For example:
 - [SCRIPT_NAME](https://www.pgadmin.org/docs/pgadmin4/8.4/container_deployment.html#reverse-proxying:~:text=The%20SCRIPT_NAME%20environment%20variable%20has%20been%20set%20to%20tell%20the%20container%20it%20is%20being%20hosted%20under%20a%20subdirectory%20(in%20the%20same%20way%20as%20the%20X%2DScript%2DName%20header%20is%20used%20with%20Nginx)%2C%20and%20a%20label%20has%20been%20added%20to%20tell%20Traefik%20to%20route%20requests%20under%20the%20subdirectory%20to%20this%20container.) for pgAdmin 4
 - [QUERY_BASE_PATH](https://www.jaegertracing.io/docs/1.55/deployment/#ui-base-path) for Jaeger

ZPain8464:
Apologies for not responding to this sooner. Yes, more examples and information in general would be helpful. 

Side note: the pgAdmin 4 link 404s. Is [this config example](https://www.pgadmin.org/docs/pgadmin4/8.9/container_deployment.html#pgadmin-x-forwarded-configuration) along the lines of what you mean?

kenjenkins:
[...]

PR #1641, opened 2024-10-07
add warning to path rewriting settings

kenjenkins:
We've had a few issue reports suggesting that users may expect Pomerium to rewrite URLs in response bodies based on the path rewriting settings. Add a caution note to explain that Pomerium does not do this.

Related issue:
- https://github.com/pomerium/documentation/issues/1299

### DIFF
--- a/content/docs/reference/routes/path-rewriting.mdx
+++ b/content/docs/reference/routes/path-rewriting.mdx
@@ -28,7 +28,7 @@ In general, if you want to expose an upstream service under some URL path prefix
 <details>
 <summary>Example: Grafana</summary>
 
-If you host a Grafana instance at `https://example.com/grafana/`, using a Pomerium route with the settings
+If you host a Grafana instance at `https://example.com/grafana/`, using a Pomerium route with the following settings:
 
 ```
   prefix: /grafana/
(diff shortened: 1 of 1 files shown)


content/docs/reference/routes/path-rewriting.mdx
```diff
@@ -19,6 +19,30 @@ This reference covers all of Pomerium's **Path Rewriting Settings**:
...
+:::caution
+
+These settings affect only the _requests_ sent to an upstream service. Pomerium will not rewrite responses from an upstream service based on these settings. This includes any link URLs or paths to JavaScript or CSS assets.
+
+In general, if you want to expose an upstream service under some URL path prefix, you will _also_ need to configure the service so it is aware of this path prefix. Otherwise, absolute URLs in the response may be incorrect and the service may not function correctly.
+
+<details>
+<summary>Example: Grafana</summary>
+
+If you host a Grafana instance at `https://example.com/grafana/`, using a Pomerium route with the settings
```
ZPain8464, line 31:
```suggestion
If you host a Grafana instance at `https://example.com/grafana/`, using a Pomerium route with the following settings:
```

ZPain8464 (approved):

Issue #1638, opened 2024-10-03
Clarify global timeouts settings


kenjenkins:
Page: https://www.pomerium.com/docs/reference/global-timeouts

## What's incorrect or missing

1. The description for "Read Timeout" is confusing to me. What does it mean for "the client to receive an entire HTTP request stream"? (The client is sending the request, not receiving it.)
2. We should document any timeouts that can be disabled by setting to 0. ("Write Timeout" lists this as the default value but does not explicitly state how this behaves.)

## What's the resolution?

1. It looks like the "Read Timeout" setting controls the Envoy HTTP connection manager [`request_timeout`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-request-timeout) setting. (see [config/envoyconfig/listeners.go#L322](https://github.com/pomerium/pomerium/blob/410354bc00c430aaf230e29715c4c7f723e807fe/config/envoyconfig/listeners.go#L322)) The Envoy docs have a more precise definition of how this timeout is applied.
2. I think all three of these timeouts can be disabled by setting them to 0, but we should verify this behavior.

wasaga:
https://www.pomerium.com/docs/internals/connection#timeouts

kenjenkins:
Thanks Denis, would be good to include a similar diagram here. One wrinkle: I think these timeouts are enforced by Envoy, not by the Go net/http package, so we should probably double-check that they are measured the same way and tweak that diagram if not.

desimone:
I believe I copied that diagram, a very long time ago, from fillipo's blog post on go's http timeouts. 

https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

ZPain8464:
No, I reconstructed this diagram in Miro.

kenjenkins:
Next steps for this issue:
- Consider adding a diagram of the request/response flow, and indicate what each timeout measures (maybe this should even show both the downstream request/response and the upstream request/response).

PR #1643, opened 2024-10-08
revise global timeouts explanation

kenjenkins:
Update the Read Timeout and Write Timeout descriptions. Call out the special behavior of the 0 value for all three. Add a sentence to discourage turning off the Read Timeout.

Related issue:
- https://github.com/pomerium/documentation/issues/1638


ZPain8464 (approved):

ZPain8464:
Thanks for updating these definitions, @kenjenkins ; they make a lot more sense!

ZPain8464:
I think we should backport these changes as well.

PR #1644, opened 2024-10-09
add kubernetes service account token file

calebdoxsey:
For https://github.com/pomerium/pomerium-console/issues/4255

Break this setting up into separate settings.


ZPain8464 (approved):

ZPain8464:
@calebdoxsey looks like this is in Core and Enterprise, correct? Do we have a docs ticket? If not, I'll add one.

Issue #1632, opened 2024-10-03
Documentation: Core to Zero import feature


ZPain8464:
This issue tracks documentation for the Core to Zero import feature. 

Deliverables include:
- [x] Guide: How to migrate to Pomerium Zero from Core OSS
- [x] Update in-app and API documentation with details of the import feature.
    - No need for changes here. The API was added to an internal api, not the public Zero api.
- [x] Reference the import process in Zero onboarding guides and quickstarts.

PR #1646, opened 2024-10-15
Documents Core to Zero import utility feature

ZPain8464:
This PR: 
- Adds a Core to Zero import page to the Zero sidebar
- Adds a Core to Zero section to the Zero Install page
- Creates an admonition callout for the new feature
- Adds this admonition to the Core Quickstart, Kubernetes Quickstart, Zero Quickstart, and the Conclusion page of the Core Courses.

Resolves https://github.com/pomerium/documentation/issues/1632


content/docs/zero/import.md
```diff
@@ -0,0 +1,125 @@
...
+
+This feature automates the process of importing your existing configuration so you don’t have to.
+
+## Core to Zero import definitions
+
+- **Entity:** any route, policy, certificate, or self-hosted domain defined in a Pomerium configuration.
+- **Replica:** a separate Pomerium configuration that shares a storage backend with other replicas running in the same deployment.
+- **Quota:** the amount of entities and replicas you can define in a cluster for free.
+
+## Supported settings
```
rjbeers, line 31: What is the value of providing this information? I don't think we want to bother the reader to go and confirm if they are using these configurations, since our import process will tell them later. Maybe this is useful at the bottom of the page as supplemental info?

content/docs/zero/import.md
```diff
@@ -0,0 +1,125 @@
...
+### Before you start
+
+To complete this guide, you need a:
+
+- [Pomerium Zero](https://console.pomerium.app/create-account) account
+- [Pomerium Core](/docs/core/quickstart) configuration
+
+:::note
+
+This guide assumes Docker as the deployment environment.
```
rjbeers, line 76: Let's call out that however you are using Core today, you can import the configuration into Zero.

rjbeers (changes requested):
I'm confused by the callouts under Next Steps.

Also we should address the case where the Core user found the import capability in our docs and is not a user of Zero yet. That reader will want to learn about the value of Zero, import capability, and how to get started with zero by importing what they have in core to see if they prefer using Zero.

content/docs/zero/import.md
ZPain8464: I'll move Quotas and Supported settings to a subsection at the bottom of the page called "Additional resources". Let me know your thoughts here; otherwise, we can remove this section entirely.

content/docs/admonitions/_import-core-config-to-zero.mdx
```diff
@@ -0,0 +1,3 @@
+If you have an existing Pomerium Core configuration, you can import it into Pomerium Zero.
```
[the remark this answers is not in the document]
ZPain8464: Thanks for the feedback. I'll update the admonition.

content/docs/quickstart.mdx
```diff
@@ -183,6 +184,12 @@ To see certificates in your cluster, go to the **Certificates** tab:
 
 ## Next Steps
 
-### [**Build your First Route**](/docs/courses/fundamentals/zero-build-routes)
+:::info Import Pomerium Core configuration
```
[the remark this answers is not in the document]
ZPain8464: I added an admonition to this section that explains a bit more about the feature and links to the guide. I also included tabs in the guide. The UI differs very subtly, but I'd rather include accurate screenshots of the UI based on the user's journey; that is, new Zero users or current Zero users.

Issue #1656, opened 2024-10-18
Missing code block titles


kenjenkins:
Page: various; one example: https://www.pomerium.com/docs/capabilities/hosted-authenticate-service

## What's incorrect or missing

Some of the code block titles are missing. Compare the "Configure the hosted authenticate service" code snippet between [v0.24](https://0-24-0.docs.pomerium.com/docs/capabilities/hosted-authenticate-service#configure-the-hosted-authenticate-service) and the [current version](https://www.pomerium.com/docs/capabilities/hosted-authenticate-service#configure-the-hosted-authenticate-service):

![Screenshot 2024-10-18 at 10 14 02 AM](https://github.com/user-attachments/assets/e8e572aa-581d-47ae-9515-bba281fe23d5)

![Screenshot 2024-10-18 at 10 14 08 AM](https://github.com/user-attachments/assets/506f1761-071f-4636-830d-037c41526c13)

## What's the resolution?

It looks like Docusaurus now requires quotes (`"`) around the code block title (see the example at https://docusaurus.io/docs/markdown-features/code-blocks#code-title).

I think this GitHub search query should find all the affected code blocks: https://github.com/search?q=repo%3Apomerium%2Fdocumentation+%2Ftitle%3D%5B%5E%22%5D%2F&type=code

PR #1657, opened 2024-10-18
fix code block title syntax

kenjenkins:
Docusaurus v3 appears to require double quotes around code block titles.

Related issue:
- https://github.com/pomerium/documentation/issues/1656


nhayfield (approved):
It does look better imo

PR #1658, opened 2024-10-18
k8s api access: add exec-api-version option to the command

wasaga:
newer versions of kubectl expect an api-version in the kubectl config

related: https://github.com/pomerium/internal/issues/1916


wasaga:
there seems to be more combinations of kubectl versions to account for

PR #1664, opened 2024-10-23
Update wistia links to youtube

cmo-pomerium:
as title, we're shifting away from Wistia to Youtube links so I'm doing a batch update based on the ones we had uploaded


rjbeers (commented):
- Some of the thumbnails don't seem as clean as the current ones.
- Some video thumbnails are zoomed in, cutting off the loom face in the bottom left
- Youtube adds "More Videos" overlay that wistia doesn't have

I think we should try to address these if possible, but if you feel strongly about youtube, let's merge this in the meantime.

rjbeers (approved):
Based on our discussion today 👍🏻

PR #1666, opened 2024-10-24
Fix ServiceMonitor in k8s install.md

nareddyt:
1. `app.kubernetes.io/instance` is the wrong label. Current `ServiceMonitor` config doesn't select the right pods.
2. For consistency, change the example namespace.


kralicky (approved):
Good catch, thank you!

PR #1676, opened 2024-10-25
Re-Adds code-server redirect to marketing integrations page

rjbeers:
Reverts pomerium/documentation#1649

We want this in the docs as well as the marketing integrations page. There was a miscommunication.


cmo-pomerium (approved):
LGTM, I see code-server back in the sidebar

PR #1685, opened 2024-10-31
Add JWT auth to Kubernetes access docs

kralicky:
This adds docs for configuring JWT auth using structured authentication as an alternative to impersonation headers. Also added a section on configuring RBAC for authorizing users, which is relevant for either method.


content/docs/capabilities/kubernetes-access.mdx
```diff
@@ -0,0 +1,241 @@
...
+- Global, cross resource access and audit trail
+- API server protection without the operational challenges of VPN
+- Can be hosted inside your kubernetes cluster!
+
+## How it works
+
+Pomerium supports two different methods of authenticating to the Kubernetes API server:
+
+- **Impersonation Headers**
+- **Pomerium JWT** using Structured Authentication (Kubernetes 1.30+)
```
rjbeers, line 47: Can we clarify the pros and cons to help the user choose the tab that is better for them?

e.g.
- **Impersonation Headers** works on all managed or unmanaged kubernetes clusters, but provides broad service account access
- **Pomerium JWT** relies on Structured Authentication Configuration, available in Kubernetes 1.30+ and may depend on your managed cluster provider's feature availability, but provides a more secure authorization policy

Probably you could clean those two statements up better than this.

rjbeers: Also, let's stay consistent and call this Structured Authentication Configuration rather than just Structured Authentication.

rjbeers (changes requested):
Requesting a minor change that could help the user flow through the document.

Thanks for updating the diagram!

rjbeers (approved):

PR #1686, opened 2024-10-31
Add zero helm chart install docs

kralicky:
- Add instructions for installing zero with the helm chart
- Remove the old "manual" tabs, as they aren't really necessary
- Re-order tabs to put kubernetes first

### DIFF
--- a/content/docs/zero/install.mdx
+++ b/content/docs/zero/install.mdx
@@ -68,7 +68,7 @@ Save the following configuration as `compose.yaml`:
 ```yaml title="compose.yaml"
 services:
   pomerium:
-    image: pomerium/pomerium:v0.27.2
+    image: pomerium/pomerium:v0.28.0
     ports:
       - 443:443
     restart: always
@@ -144,7 +144,7 @@ In v0.27, we updated the Kubernetes installation manifest to use a Deployment in
 To update Pomerium in Kubernetes, run the following command:
 
 ```bash
-$ kubectl apply -k github.com/pomerium/pomerium/k8s/zero\?ref=<X.Y.Z>
+$ kubectl apply -k github.com/pomerium/pomerium/k8s/zero\?ref=0.28.0
 ```
 
 </TabItem>
@@ -156,7 +156,7 @@ To update Pomerium in Docker:
 
 ```yaml
 pomerium:
-  image: pomerium/pomerium:<vX.Y.Z>
+  image: pomerium/pomerium:v0.28.0
 ```
 
 1. Run the following command:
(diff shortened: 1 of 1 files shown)


content/docs/zero/install.mdx
```diff
@@ -25,18 +25,21 @@ If you haven't signed up for Pomerium Zero, you can [**create an account**](http
...
 
-Install Pomerium with the following shell script:
+To install Pomerium with Helm, run the following command:
 
 ```bash
-curl https://console.pomerium.app/install.bash | env POMERIUM_ZERO_TOKEN=<CLUSTER_TOKEN> bash -s install
+helm install pomerium-zero oci://docker.io/pomerium/pomerium-zero \
+  --set pomeriumZeroToken=<CLUSTER_TOKEN> \
+  --namespace pomerium-zero \
+  --create-namespace
```
wasaga, line 36: can we make those two defaults so that only cluster token would be a required arg

content/docs/zero/install.mdx
```diff
@@ -25,18 +25,21 @@ If you haven't signed up for Pomerium Zero, you can [**create an account**](http
...
 <Tabs>
-<TabItem value="linux" label="Linux">
+<TabItem value="kubernetes-helm" label="Kubernetes (Helm)">
 
-Install Pomerium with the following shell script:
+To install Pomerium with Helm, run the following command:
 
 ```bash
-curl https://console.pomerium.app/install.bash | env POMERIUM_ZERO_TOKEN=<CLUSTER_TOKEN> bash -s install
+helm install pomerium-zero oci://docker.io/pomerium/pomerium-zero \
```
wasaga, line 33:
```
$ helm install pomerium-zero oci://docker.io/pomerium/pomerium-zero
Error: INSTALLATION FAILED: GET "https://registry-1.docker.io/v2/pomerium/pomerium-zero/tags/list": unexpected status code 401: unauthorized: authentication required
```

content/docs/zero/install.mdx
```diff
@@ -101,52 +104,36 @@ docker compose up -d
...
+<TabItem value="kubernetes-helm" label="Kubernetes (Helm)">
 
-1. Install the latest version of Pomerium:
+To update Pomerium with Helm, run the following command:
 
 ```bash
-$ sudo yum update pomerium
+helm upgrade pomerium-zero oci://docker.io/pomerium/pomerium-zero \
+  --namespace pomerium-zero \
+  --version <X.Y.Z>
```
wasaga, line 132: what is `X.Y.Z` in this context? how would user know that? is there a way to just upgrade to latest?

content/docs/zero/install.mdx
kralicky: You're not supposed to put a namespace in the chart templates, since it's specified from this arg. I don't think helm will create the namespace unless you add --create-namespace either

content/docs/zero/install.mdx
kralicky: chart should be available now

content/docs/zero/install.mdx
kralicky: This is how the existing docs on this page wrote the version string, I assumed it was something easy to find/replace for releases? not sure

wasaga (approved):

PR #1703, opened 2024-11-19
settings: add http3 codec type

calebdoxsey:
Document https://github.com/pomerium/internal/issues/1999.

### DIFF
--- a/content/docs/reference/codec-type.mdx
+++ b/content/docs/reference/codec-type.mdx
@@ -58,3 +58,53 @@ Kubernetes does not support `codec_type`.
 | `http1` | Only support `http1` |
 | `http2` | Only support `http2` |
 | `http3` | Starts a both UDP and TCP listener on [address](/docs/reference/address). The UDP listener is configured to use QUIC with HTTP/3. The TCP listener is configured to use `auto` as the codec type, and an `Alt-Svc` header is added to all responses indicating `h3` is available (Available since v0.29) |
+
+## Protocol Details
+
+### HTTP/1.1 (`http1`)
+
+HTTP/1.1 is the traditional protocol for web communication, operating over TCP. It supports basic request-response interactions but lacks advanced features like multiplexing, which can lead to inefficiencies such as head-of-line blocking.
+
+### HTTP/2 (`http2`)
+
+HTTP/2 enhances performance over HTTP/1.1 by introducing:
+
+- **Multiplexing**: Allows multiple requests and responses concurrently over a single connection.
+- **Header Compression**: Reduces overhead by compressing HTTP headers.
+- **Server Push**: Enables servers to send resources proactively to clients.
+
+These features improve page load times and resource utilization, especially for applications with multiple simultaneous requests.
+
+### HTTP/3 (`http3`)
+
+HTTP/3 is the latest version of the HTTP protocol, utilizing QUIC over UDP. Benefits include:
+
+- **Reduced Latency**: Faster connection establishment with 0-RTT (zero round-trip time) support.
+- **Improved Performance**: Better handling of packet loss and network congestion.
+- **Connection Migration**: Maintains connections even when network changes occur (e.g., switching from Wi-Fi to cellular).
+
+**Considerations**:
+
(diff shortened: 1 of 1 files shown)


content/docs/reference/codec-type.mdx
```diff
@@ -47,7 +45,16 @@ CODEC_TYPE=http2
...
 </Tabs>
+
+### Options
+
+| **Options** | **Description** |
+| :-- | :-- |
+| `auto` | Either `http1` or `http2` will be determined via TLS ALPN or protocol inference |
+| `http1` | Only support `http1` |
+| `http2` | Only support `http2` |
+| `http3` | Starts a both UDP and TCP listener on [address](/docs/reference/address). The UDP listener is configured to use QUIC with HTTP/3. The TCP listener is configured to use `auto` as the codec type, and an `Alt-Svc` header is added to all responses indicating `h3` is available (Available since v0.29) |
```
desimone, line 60:
```suggestion
| `http3` | Starts a both UDP and TCP listener on [address](/docs/reference/address). The UDP listener is configured to use QUIC with HTTP/3. The TCP listener is configured to use `auto` as the codec type, and an `Alt-Svc` header is added to all responses indicating `h3` is available (Available since v0.29) |

## Protocol Details

### HTTP/1.1 (`http1`)

HTTP/1.1 is the traditional protocol for web communication, operating over TCP. It supports basic request-response interactions but lacks advanced features like multiplexing, which can lead to inefficiencies such as head-of-line blocking.

### HTTP/2 (`http2`)

HTTP/2 enhances performance over HTTP/1.1 by introducing:

- **Multiplexing**: Allows multiple requests and responses concurrently over a single connection.
- **Header Compression**: Reduces overhead by compressing HTTP headers.
- **Server Push**: Enables servers to send resources proactively to clients.

These features improve page load times and resource utilization, especially for applications with multiple simultaneous requests.

### HTTP/3 (`http3`)

HTTP/3 is the latest version of the HTTP protocol, utilizing QUIC over UDP. Benefits include:

- **Reduced Latency**: Faster connection establishment with 0-RTT (zero round-trip time) support.
- **Improved Performance**: Better handling of packet loss and network congestion.
- **Connection Migration**: Maintains connections even when network changes occur (e.g., switching from Wi-Fi to cellular).

**Considerations**:

- **Network Compatibility**: Ensure that firewalls, routers, and network devices allow UDP traffic on the necessary ports. Some networks may block UDP traffic, which can prevent HTTP/3 connections from being established.
- **Fallback Mechanism**: When HTTP/3 is enabled, Pomerium also provides a TCP listener that supports `http1` and `http2` protocols. This allows clients that cannot use HTTP/3 (due to network restrictions or lack of support) to fall back to earlier protocol versions.
- **Client Support**: Only clients that support HTTP/3 and QUIC can leverage these benefits. Older clients will continue to use HTTP/1.1 or HTTP/2.

:::enterprise

Before you can generate device registration links for users within your directory, you must sync your directory data first.

See [**Directory Sync**](/docs/capabilities/directory-sync) for more information.

:::warning

Because HTTP/3 relies on QUIC over UDP, it may be blocked by some network environments that restrict UDP traffic. Ensure that your deployment environment permits UDP on the necessary ports, and be prepared for clients in restrictive networks to fall back to HTTP/1.1 or HTTP/2.

:::

When `codec_type` is set to `http3`, Pomerium operates as follows:

- **UDP Listener**: Listens for QUIC connections to serve HTTP/3 traffic.
- **TCP Listener**: Continues to support `http1` and `http2` based on the `auto` codec type.
- **Alt-Svc Header**: Adds an `Alt-Svc` header to HTTP responses, informing clients that HTTP/3 is available.

This configuration allows clients that support HTTP/3 to upgrade their connections while maintaining compatibility with clients using earlier protocol versions.
```

Let me know if that additional context is helpful and -- more importantly -- correct. 

I got these details from reading:

- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http3
- https://baptistout.net/posts/upgrade-envoy-http3/

Issue #214, opened 2022-12-19
docs.pomerium.com should redirect to pomerium.com/docs


wasaga:
docs.pomerium.com should redirect to pomerium.com/docs

desimone:
It's difficult to redirect (netlify controlls the subdomain) but the content will be equivalent; only `pomerium.com/docs` will have the canonical tag which should point folks in the right direction. 

I've added a `_redirect` entry but I doubt it will work.

desimone:
This might need to be fixed at the DNS (netlify) or CDN / LB layer (GCP). It's... tricky.

ZPain8464:
This one redirects to docs.pomerium.com/docs currently

desimone:
I can't figure this out, unfortunately.

rjbeers:
> docs.pomerium.com should redirect to pomerium.com/docs

Why do we want this behavior? docs.pomerium.com resolves to the nearest netlify server, with coverage around the world. www.pomerium.com/docs resolves to a GCP LB which then proxies to the closest netlify server. Do we want that funnel through GCP for some reason?

PR #1709, opened 2024-11-19
redirect docs.pomerium.com to www.pomerium.com

kenjenkins:
Related issue:
- https://github.com/pomerium/documentation/issues/214

Unfortunately I don't know if there's any way to test this short of deploying it to production.


rjbeers (commented):
I believe this will create an infinite loop, because www.pomerium.com/docs proxies to https://docs.pomerium.com/docs.

We need to update our infrastructure proxying logic to proxy to pomerium-docs.netlify.app, and then we can have these redirects in place to send traffic to www

PR #1712, opened 2024-12-10
oidc/azure: add a note to get v2 tokens

wasaga:
Adds a note to obtain V2 tokens in Microsoft Entra.

Related: https://linear.app/pomerium/issue/ENG-1736/coreoidc-microsoft-entra-returns-v1-tokens-which-have-limited-set-of

### DIFF
--- a/content/docs/identity-providers/azure.mdx
+++ b/content/docs/identity-providers/azure.mdx
@@ -92,7 +92,7 @@ You will use the [**Group ID**](https://docs.microsoft.com/en-us/graph/api/group
 
 :::note
 
-Pomerium uses v2.0 Entra Access Token, make sure your application manifest has [`accessTokenAcceptedVersion`](https://learn.microsoft.com/en-us/entra/identity-platform/reference-app-manifest#accesstokenacceptedversion-attribute) set to `2`. If you use Terraform to configure your Entra application, set [`requested_access_token_version = 2`](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/application.html#requested_access_token_version-1) of an `azuread_application`.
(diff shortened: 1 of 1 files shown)


calebdoxsey (approved):

content/docs/identity-providers/azure.mdx
```diff
@@ -90,6 +90,12 @@ You will use the [**Group ID**](https://docs.microsoft.com/en-us/graph/api/group
 
 :::
 
+:::note
+
+Pomerium uses v2.0 Entra Access Token, make sure your application manifest has [`accessTokenAcceptedVersion`](https://learn.microsoft.com/en-us/entra/identity-platform/reference-app-manifest#accesstokenacceptedversion-attribute) set to `2`. If you use Terraform to configure your Entra application, set [`requested_access_token_version = 2`](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/application.html#requested_access_token_version-1) of an `azuread_application`.
```
kenjenkins, line 95: Based on Slack discussion we think the v2 format is not a hard requirement, but we'd like to recommend using it anyway.

Maybe we could phrase this note something like this?

```suggestion
We recommend using the v2.0 access token format, which you can request by setting the application manifest [`accessTokenAcceptedVersion`](https://learn.microsoft.com/en-us/entra/identity-platform/reference-app-manifest#accesstokenacceptedversion-attribute) attribute to `2`. If you use Terraform to configure your Entra application, set [`requested_access_token_version = 2`](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/application.html#requested_access_token_version-1) in your `azuread_application`.
```

kenjenkins (approved):

PR #1728, opened 2025-01-03
Correct our support for UDP and HTTP/3

rjbeers:
This was missed in prior docs PRs related to UDP and HTTP/3 work.


rjbeers:
This page is trying to focus on HTTP requests, but I wonder if we should expand and talk about how TCP and UDP tunneled traffic is handled. Maybe that's a separate PR entirely.

calebdoxsey (approved):

PR #1779, opened 2025-02-11
ENG-1994: Link to latest 3 versions of the docs

rjbeers:
I removed the horizontal rule as well. This is personal preference. See the preview for how it looks. Fine with adding it in if others have strong opinions.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -137,12 +137,12 @@ const config = {
           dropdownActiveClassDisabled: true,
           dropdownItemsAfter: [
             {
-              to: 'https://0-28-0.docs.pomerium.com/docs',
-              label: 'v0.28 (latest)',
+              to: 'https://0-27-0.docs.pomerium.com/docs',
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -137,12 +137,12 @@ const config = {
           dropdownActiveClassDisabled: true,
           dropdownItemsAfter: [
             {
-              to: 'https://0-28-0.docs.pomerium.com/docs',
-              label: 'v0.28 (latest)',
+              to: 'https://0-27-0.docs.pomerium.com/docs',
```
kenjenkins, line 140: I think we should keep the link to the latest released version. Here it skips from "vNext" (upcoming v0.29) to v0.27.

rjbeers: Yes. I was going to do a follow-up PR to prep the main branch for a 0.29, but figured this is the real change we want into 0.28. Anyway, I think you made the right change, so I'll close this.

PR #1781, opened 2025-02-11
fix version dropdown

kenjenkins:
Update the version dropdown so it contains v0.28, v0.27, and v0.26.

Related issue:
https://linear.app/pomerium/issue/ENG-1994/docs-live-site-version-dropdown-contains-two-v028-links


nickytonline (approved):
Tested the deploy preview and looks good and opens the correct versions of docs.

PR #1782, opened 2025-02-12
add more redirects

kenjenkins:
I happened across a broken external link to the CORS guide, so I figured we should add a redirect for that. I also exported a recent list of 404s from Google Search Console and added redirects for most of those (some of them I couldn't find on the current site).


nickytonline (approved):
Took the deploy preview for a spin. Looks good!

PR #1796, opened 2025-03-05
bearer-token-format: add az example, document enterprise options

calebdoxsey:
For [ENG-2085](https://linear.app/pomerium/issue/ENG-2085/document-all-possible-values-of-bearer-token-format)

### DIFF
--- a/content/docs/reference/bearer-token-format.mdx
+++ b/content/docs/reference/bearer-token-format.mdx
@@ -63,11 +63,19 @@ bearer_token_format: idp_access_token
 BEARER_TOKEN_FORMAT=idp_access_token
 ```
 
+#### Microsoft Entra
+
+The `az` CLI can be used to get an access-token:
+
+```bash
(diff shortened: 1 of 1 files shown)


content/docs/reference/bearer-token-format.mdx
```diff
@@ -76,6 +84,14 @@ Set **Bearer Token Format** under **Proxy** settings in the Console:
 
 ![Set bearer token format in the Console](./img/global-settings/bearer-token-format.png)
 
+### Options
+
+- _Unset_: At the route level, use the global setting. At the global level, use "Default".
+- "Unknown": Same as "Default".
```
rjbeers, line 90: Seems like we shouldn't have this value available in the UI, but I suppose this isn't the only place we have this, so let's leave it and document it for now and discuss how we can address these Unknowns where they exist.

rjbeers (approved):

PR #1798, opened 2025-03-06
adds the idp keycloak help info - ENG-2107

nhayfield:


rjbeers (commented):
Where are entries that have `"enterpriseOnly": true` used?

nhayfield:
> Where are entries that have `"enterpriseOnly": true` used?

They are used in enterprise for a few places. The flag prevents them from showing up in the docs reference table.

rjbeers (approved):

PR #1803, opened 2025-03-25
OPE-117 add _headers to not index

gaurdro:
This adds a default header that marks docs as not to be indexed.   With a follow up, we'll want to make sure we add process to make sure the current current docs are allowed to be indexed. 

From the preview deployment a curl shows we have the desired header 
Screenshot 2025-03-24 at 8 28 27 PM

### DIFF
--- a/_headers
+++ b/_headers
@@ -1,0 +1,2 @@
+/*
+  X-Robots-Tag: noindex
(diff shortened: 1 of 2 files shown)


desimone:
RE: follow up, we should make sure `X-Robots-Tag: all` is set on the current active docs.

desimone (approved):

PR #1804, opened 2025-03-25
ENG-2184 Update docs to use signing_key_file

tsullyo:
- **Updated jwt verification guide for Core to use file-based signing key**
- **Updated other instances of using a hard-coded signing key**

### DIFF
--- a/content/docs/get-started/fundamentals/core/jwt-verification.md
+++ b/content/docs/get-started/fundamentals/core/jwt-verification.md
@@ -109,7 +109,7 @@ Route settings must be attached to the route itself and will only apply to that 
 
 Configuring identity verification is a great example of how you should organize these settings in your configuration files:
 
-- **Signing Key** is typically a global-level setting. Any time a user sends a request to an upstream service, Pomerium will look for a signing key.
+- **Signing Key** is a global-level setting. Pomerium will use this key to sign all of the JWTs that it sends to an upstream service.
 - **Pass Identity Headers** can be configured as a global or route-specific setting. It tells Pomerium to send the signed header to a certain route.
 
 In this way, global and route level settings allow you to fine tune your configuration to suit a service's use case.


content/docs/get-started/fundamentals/core/jwt-verification.md
```diff
@@ -99,25 +99,27 @@ There are two settings that you need to configure to implement identity verifica
...
+Global settings are settings that will apply to ALL resources protected by Pomerium. Global settings are added _outside_ of your route and policy blocks.
 
-Route settings must be attached to the route itself.
+Route settings must be attached to the route itself and will only apply to that specific route/policy block.
 
 Configuring identity verification is a great example of how you should organize these settings in your configuration files:
 
-- **Signing Key** is a global-level setting. Any time a user sends a request to an upstream service, Pomerium will look for a signing key.
-- **Pass Identity Headers** is a route-level setting. It tells Pomerium to send the signed header to a certain route.
+- **Signing Key** is typically a global-level setting. Any time a user sends a request to an upstream service, Pomerium will look for a signing key.
```
kenjenkins, line 112: I don't think "typically" is accurate here. The I believe Pomerium's [Signing Key](https://www.pomerium.com/docs/reference/signing-key) option is only available as a global setting.

kenjenkins, line 112: side note: the existing phrasing "Any time a user sends a request to an upstream service, Pomerium will look for a signing key" seems a little off. Strictly speaking, the user does not make a request to an upstream service, but rather Pomerium makes that request on behalf of the user. 

Maybe something more like this? "Pomerium will use this key to sign all of the JWTs that it sends to an upstream service"

kenjenkins (approved):
Thanks, I think this is a good change!

content/docs/get-started/fundamentals/core/jwt-verification.md
tsullyo: Agreed. I'll edit and update the PR

tsullyo: Agreed. I'll edit and update the PR

tsullyo:
Ughhh. Now I'm in rebase hell. Do we have a standardized git strategy to follow?

tsullyo:
The diff on this PR shows edits outside the scope of my intended changes. As my changes are ultimately small, I'm going to abandon this and create a new PR rather than try to untangle .

PR #1726, opened 2024-12-29
docs: restructure, clarify, consolidate existing docs

desimone:
**Summary**  
This PR reorganizes our documentation to unify repetitive content (e.g., multiple versions of identity/JWT verification, CLI usage, and certificate guides) into a more readable structure. We want newcomers to see a clear “start here” path, discover high-level product capabilities, and then explore advanced references once they’re ready for production. No changes have been made to our existing reference docs or guides.

### Improved Onboarding Flow
- **Fundamentals** A step-by-step tutorial for new users (Quickstart, policy building, SSO, advanced routes).
- **Capabilities** High-level feature topics (authorization, routing, branding, etc.) that highlight Pomerium’s value proposition and show how each feature solves user problems.
- **Internals**  Advanced deep-dives (architecture, zero trust, connection details) for production or experienced users wanting best practices and deeper context.

### Consolidated Overlapping Docs (many others)
- **Identity & JWT Verification** Merged scattered pages (old “JWT Verification,” separate identity provider docs) into a single “Getting Users’ Identity” doc.
- **CLI/Desktop** Unified multiple how-to guides for TCP/UDP tunneling into a single “Clients for TCP & UDP” doc, clarifying both CLI and Desktop usage.
- **Certificates** Streamlined references on certificate usage and mTLS (previously found in separate guides) into one “Certificates & TLS” doc.

### Docs Platform Improvements
- Removed some older version references from the sidebar dropdown.
- Fixed syntax highlighting issues.
- Fixed a hack (e.g., a custom webpack plugin snippet) in `docusaurus.config.js` for future use.

### Follow-Up Work
- [x] Ensure all old links have proper redirects (hell)
- [x] Update the overview page’s flow diagram; using a mermaid placeholder for now 
- [ ] Further unify concepts and glossary items  
- [ ] On the “Get Started” page, add tabbed instructions for both Core and Zero so users drop into the correct fundamentals  
- [ ] Consider reorganizing reference docs for better navigation  
- [ ] Fix Algolia’s search results (the new consolidation should help)  
- [ ] Move the JIT access doc into the PPL section and expand PPL coverage  
- [ ] Overhaul the “Deployment” section for clarity—guide GCP or K8s Ingress use cases, unify Kubernetes docs more effectively  
- [ ] Decide whether to keep or update the old “High Availability” doc (merged with “config”)  
- [ ] Add Autocert details to the automatic domain registration doc  
- [ ] Reference Envoy usage in “Proxying & Routing” docs  
[...]

PR #1860, opened 2025-04-22
(docs) refactor take 1

wrmedford:


desimone:
I know this is still a draft, but can we break these down into smaller, atomic changes? Let's also make sure to include redirects for any proposed URL updates. 

[speaking from experience], 😱  I made similar large-scale changes, and they were tough to migrate(https://github.com/pomerium/documentation/pull/1726) and find the 404s. 

There's still a lot of unfinished changes from that PR too

PR #1886, opened 2025-06-16
add circuit breaker thresholds

calebdoxsey:
Document circuit breaker thresholds.

For [ENG-2477](https://linear.app/pomerium/issue/ENG-2477/documentation-document-circuit-breaker-settings)

### DIFF
--- a/content/docs/reference/circuit-breaker-thresholds.mdx
+++ b/content/docs/reference/circuit-breaker-thresholds.mdx
@@ -23,7 +23,7 @@ In Pomerium requests will automatically fail when circuit breaker thresholds are
 - Max Connections: The maximum number of connections that Pomerium will establish to all hosts in a route's upstream cluster. Default is `1024`.
 - Max Pending Requests: The maximum number of requests that will be queued while waiting for a ready connection pool connection. Default is `1024`.
 - Max Requests: The maximum number of requests that can be outstanding to all route upstream hosts in a cluster at any given time. Default is `1024`.
-- Max Retries: The maximum number of retries that can be outstanding to all hosts in a route upstream cluster at any given time. Default is `3`.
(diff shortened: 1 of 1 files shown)


content/docs/reference/circuit-breaker-thresholds.mdx
```diff
@@ -0,0 +1,88 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Circuit Breaker Thresholds
+
+## Summary
+
+In Pomerium requests will automatically fail when circuit breaker thresholds are reached. The following thresholds are available:
+
+- Max Connections: The maximum number of connections that Pomerium will establish to all hosts in an upstream cluster. Default is `1024`.
```
kenjenkins, line 23: Should we try to explain what constitutes an "upstream cluster"? I think this is roughly "per route", is that right? Like, if I define two routes with the same set of `to` URLs, the connection limit is still per-route rather than shared between the routes?

content/docs/reference/circuit-breaker-thresholds.mdx
```diff
@@ -0,0 +1,88 @@
...
+- Max Pending Requests: The maximum number of requests that will be queued while waiting for a ready connection pool connection. Default is `1024`.
+- Max Requests: The maximum number of requests that can be outstanding to all hosts in a cluster at any given time. Default is `1024`.
+- Max Retries: The maximum number of retries that can be outstanding to all hosts in a cluster at any given time. Default is `3`.
```
kenjenkins, line 26: Do we enable automatic retries anywhere? I'm not entirely sure when this limit would be relevant.

kenjenkins (approved):

PR #1927, opened 2025-07-18
fix: llms-txt uses markdown files now

nickytonline:
Backports #1921 to the 0.30 release docs.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -232,7 +232,6 @@
     "lifecycles",
     "llms",
     "Llms",
-    "TTL",
     "relogin",
     "quantiles",
     "pgxpool",
--- a/cspell.json
+++ b/cspell.json
@@ -231,7 +231,15 @@
     "HPKE",
     "lifecycles",
     "llms",
-    "Llms"
+    "Llms",
+    "TTL",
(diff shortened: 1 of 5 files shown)


cspell.json
```diff
@@ -231,7 +231,15 @@
     "HPKE",
     "lifecycles",
     "llms",
-    "Llms"
+    "Llms",
+    "TTL",
```
kenjenkins, line 235: "TTL" wasn't part of #1921, was it?

nickytonline: I think something else in main brought it in because my PR was based off of main. I ran the instructions from the failed backport though. I can remove it.

package.json
```diff
@@ -42,7 +42,12 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-mailchimp-subscribe": "^2.1.3",
-    "redocusaurus": "^2.2.3"
+    "redocusaurus": "^2.5.0",
```
kenjenkins, line 45: This change was also not part of #1921. If you keep this update to redocusaurus you should update yarn.lock to match.

PR #1928, opened 2025-07-19
add ssh to the 'From' URL reference

kenjenkins:
Update the list of supported URL schemes on the From URL reference page.

Related to https://linear.app/pomerium/issue/ENG-2040/document-native-ssh-access.


nickytonline (approved):
🚢

rjbeers (approved):
Good catch. We can merge this, but we should also fix the table on this page to include `ssh` as a valid scheme.

Issue #1924, opened 2025-07-18
llms.txt quickstart page tile is incorrect


nickytonline:
See https://main.docs.pomerium.com/llms.txt. Notice the first link in Getting Started is

```
- [cSpell:ignore thisisunsafe, genkey, noout](content/docs/get-started/quickstart.md): Learn how to install and run Pomerium Zero in a Docker container.
```

My hunch is `cSpell` etc. is being pulled in in an HTML comment potentially. Here's the current quickstart.md. The issue seems to be there is no title in the frontmatter.

```markdown
---
# cSpell:ignore thisisunsafe, genkey, noout

lang: en-US
sidebar_label: Quickstart
pagination_prev: null
pagination_next: null
description: Learn how to install and run Pomerium Zero in a Docker container.
keywords:
  [
    pomerium,
    identity access proxy,
    oidc,
    docker,
    reverse proxy,
    containers,
    identity aware proxy,
    quickstart pomerium,
    docker,
    docker-compose,
  ]
---

import Tabs from '@theme/Tabs';
import TabItem from '@theme/TabItem';

# Pomerium Zero Quickstart

The Zero Quickstart shows you how to install and run Pomerium Zero in a Docker container.

## Before you start

- Install [Docker](https://docs.docker.com/install/) and [Docker Compose](https://docs.docker.com/compose/install/)
- Create a (free) [**Pomerium Zero account**](https://console.pomerium.app/create-account)

## Get Pomerium Zero configuration

After you create an account, you'll be directed to the Zero onboarding screen. In the onboarding screen, select the **Docker** tab and copy the Docker Compose configuration.

:::note Import an existing Pomerium Core configuration

If you are using Pomerium Zero for the first time, but have an existing Pomerium Core configuration, you can import it into Pomerium Zero by following the prompts in the Zero onboarding flow.

See [**Import Core Config to Zero**](/docs/deploy/cloud/import#how-to-import-your-core-configuration) for more information.

:::

(If you selected **Finish** before copying the Docker configuration, we've provided a copy below.)

Paste the configuration in a `compose.yaml` file:

```yaml title="compose.yaml" {8,15} showLineNumbers
services:
  pomerium:
    image: pomerium/pomerium:latest
    ports:
      - 443:443
    restart: always
    environment:
      POMERIUM_ZERO_TOKEN: <CLUSTER_TOKEN>
      XDG_CACHE_HOME: /var/cache
    volumes:
      - pomerium-cache:/var/cache
    networks:
[...]

PR #1931, opened 2025-07-20
fix: remove YAML frontmatter from md files for cleaner llms.txt

nickytonline:
llms.txt is correctly referencing markdown content now, but the frontmatter of the markdown files in build artifacts was retained. This can cause issues with LLMs reading the markdown files (see recommended format https://llmstxt.org/#format). As well, the what I thought was a redundant title is not, since we're stripping the frontmatter, unlike when the actual pages render, so the quickstart title is back.

An example in the deploy of frontmatter stripped, but feel free to check out other files referenced in https://deploy-preview-1931--pomerium-docs.netlify.app/llms.txt:

- https://deploy-preview-1931--pomerium-docs.netlify.app/content/docs/get-started/quickstart.md

Relates to #1924

### DIFF
--- a/content/docs/get-started/quickstart.mdx
+++ b/content/docs/get-started/quickstart.mdx
@@ -23,6 +23,8 @@ keywords:
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
+# Pomerium Zero Quickstart
+
 The Zero Quickstart shows you how to install and run Pomerium Zero in a Docker container.
 
 ## Before you start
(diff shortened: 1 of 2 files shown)


kenjenkins (approved):
Unrelated to this PR, but it looks like there's also a number of broken links where the linked page starts with an underscore?

e.g. in the Other section:
```
- [_import Core Config To Zero](content/docs/admonitions/_import-core-config-to-zero.md)
- [_generate Recovery Token](content/docs/admonitions/_generate-recovery-token.md)
- [_handle Self Signed Certificate Warning](content/docs/admonitions/_handle-self-signed-certificate-warning.md)
```

nickytonline:
> _import-core-config-to-zero.md

These look like file imports which shouldn't be a part or the list of files in the llms.txt. I'll create a separate issue for that.

PR #1953, opened 2025-07-31
force upgrade of webpack-dev-server to >= 5.2.1

kenjenkins:
This should resolve a pair of dependabot security alerts on webpack-dev-server, which is currently pinned to a vulnerable version via a transitive dependency from docusaurus.


rjbeers (approved):
Preview looks good to me. Let's ship this for now.

PR #1990, opened 2025-09-08
pre-commit: update, clarify dev setup, and cleanup

rjbeers:
We use the python pre-commit system in this repo with a separate github action for it, but the configuration was outdated and I'm not sure any developers were setting it up locally, so this clarifies the installation requirements and updates related configs.

I also removed the cspell npm scripts, because you don't run them with yarn/npm.

---

Maybe I'm missing something, but as far as I can tell, we should all be following the steps I just added to the repo and relying on the pre-commit automation that matches CI. If we want to deviate from that, we could do something more native to npm/yarn with npm scripts and add the cspell dependency and maybe husky for pre-commit script management. I figured this PR requires initial setup, but then should match CI, so went with that for now. Open to discuss.


README.md
```diff
@@ -14,6 +14,27 @@ The steps below detail the installation of this site locally for development.
 yarn
 ```
 
+### Pre-commit Setup
+
+This project uses pre-commit hooks to ensure code quality. To set up pre-commit:
+
+1. Install pre-commit:
```
nickytonline, line 21: I think the friction here is it doesn't happen automatically on npm install. I definitely have no problem installing this, just wasn't aware of it.

That said since this is a JS project, I've typically used [Husky](https://typicode.github.io/husky/), which installs via npm, so one less thing for an internal or external contributor to think about and it will run without any additional setup.

rjbeers: Yeah, I agree, and husky + a cspell npm dependency was my original intent until I realized what was already setup. If we add husky based hooks and the cspell npm dependency, i'd prefer to remove the existing pre-commit stuff, so the cspell versions are consistent. Then, would netlify would run the linting though. I do like the separate lint checks, so we get a preview regardless of linting failures.

PR #1993, opened 2025-09-15
docs: add health check configuration

alexandreLamarre:
[ENG-2290](https://linear.app/pomerium/issue/ENG-2290/document-healthcheck-with-graceful-teardown)

### DIFF
--- a/content/docs/internals/health-checks.mdx
+++ b/content/docs/internals/health-checks.mdx
@@ -25,7 +25,7 @@ Health checks allow you to automatically monitor Pomerium instances and restart 
 
 In Pomerium, these health checks will report on the status of:
 
-- Readiness and health of the embedded envoy instance
+- Readiness and health of the embedded Envoy instance
 - Health of the configuration syncing mechanisms
 - Health of the configuration management server
 - Storage backend health, where applicable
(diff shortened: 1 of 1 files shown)


content/docs/internals/health-checks.mdx
```diff
@@ -0,0 +1,280 @@
...
+      periodSeconds: 60
+      failureThreshold: 5
+```
+
+### Startup
+
+The startup endpoint `/startupz` waits for all components to report ready - in practice the latest version of the configuration should be synced before starting.
+
+### Readiness
+
```
rjbeers, line 76: Missing readiness content

content/docs/internals/health-checks.mdx
```diff
@@ -0,0 +1,280 @@
...
+### Liveness
+
+The `/healthz` endpoint
```
rjbeers, line 79: Missing liveness content

content/docs/internals/health-checks.mdx
```diff
@@ -0,0 +1,280 @@
...
+          - a
+          - $(POD_IP):28080
+      initialDelaySeconds: 5
+      periodSeconds: 30
+      timeoutSeconds: 5
+```
+
+### Filters
+
+Unlike HTTP probes, the CLI provides additional flexibility through exclude filters, which let you ignore specific internal conditions reported by Pomerium when evaluating health.
```
rjbeers, line 181: How about adding a check definition section with short descriptions of what each of these checks covers, so they can figure out what they might want to filter out?

content/docs/internals/health-checks.mdx
alexandreLamarre: Looks like I accidentally git stashed those changes.

content/docs/internals/health-checks.mdx
alexandreLamarre: Yeah probably a good idea to do that, although they're internal checks so may not make intuitive sense to users, so let me take some time to think about the best way to document those

rjbeers (approved):

PR #2002, opened 2025-09-24
Miscelaneous config fixes & PPL rewording

alexandreLamarre:
- Fixes an invalid docker compose, in the fundamentals core section.
- Fixes the pomerium config for native ssh access. The `ssh_user_ca_key_file` is a string, not an array.
- Adds a redirect for an outdated link related to data storage.
- Rewords the PPL breakdown to be clearer and less technical

### DIFF
--- a/content/docs/get-started/fundamentals/core/build-policies.md
+++ b/content/docs/get-started/fundamentals/core/build-policies.md
@@ -115,8 +115,10 @@ There are four logical operators:
 
 In PPL, a criterion defines a specific condition to evaluate, such as a user’s email or device type.
 
-- Each criterion is an object where the key is the criterion name (optionally with a sub-path, delimited by `/`), and the value specifies what to match or compare.
-- The exact format of the value depends on the type of criterion being used.
+- Each criterion is represented by an object with a single key/value pair, where the key is the criterion type.
+- For some criteria the key accepts a sub-path, delimited by /. For example: claim/family_name.
+- The format of the criterion value varies depending on the criterion type.
+- Some criteria do not use their value. For example: `accept`, `reject`, and `authenticated_user`. In this case, the value can be anything.
 
 #### Matchers
 
@@ -140,11 +142,11 @@ Now that you've briefly covered PPL, let's jump into some simple examples:
 This example instructs Pomerium to only grant a user access if their email address is `example@domain.com`.
 
 ```yaml title="PPL rule"
-policy: # Policy object starts here
-  allow: # At least one action
-    and: # Logical operator
-      - email: # Criterion
-        is: example@domain.com # Value
+policy:
+  allow:
+    and:
+      - email:
+          is: example@domain.com
 ```
 
 **Example 2**: Allow access based on the domain criterion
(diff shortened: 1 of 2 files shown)


agocs (approved):

content/docs/get-started/fundamentals/core/build-policies.md
```diff
@@ -66,15 +66,39 @@ You can think of it as coded instructions to tell Pomerium how authorization dec
...
-PPL consists of **Rules**, **Actions**, **Logical Operators**, **Criteria**, and **Matchers**.
+PPL is made up of **Rules**, **Actions**, **Logical Operators**, **Criteria**, and **Matchers**.
+
+Below is a simple policy example in PPL format that allows access for the email example@domain.com. The following sections will break down each of these components in more detail.
+
+<!-- prettier-ignore-start -->
+```yaml
+# PPL Example
+policy:
+  # ────────── Rule ──────────
```
kenjenkins, line 77: If we add this annotated example here, should we remove the comments from lines 143-147?

content/docs/get-started/fundamentals/core/build-policies.md
kenjenkins, line 1: I think there's also an existing syntax problem on line 147: the `is` matcher needs to be nested under the `email` key. Would you mind fixing it while you're here?

content/docs/get-started/fundamentals/core/build-policies.md
```diff
@@ -89,7 +113,10 @@ There are four logical operators:
 
 #### Criteria
 
-Criteria in PPL are represented as an object where the key is the name and optional sub-path of the criterion, and the value changes depending on which criterion is used.
+In PPL, a criterion defines a specific condition to evaluate, such as a user’s email or device type.
+
+- Each criterion is an object where the key is the criterion name (optionally with a sub-path, delimited by `/`), and the value specifies what to match or compare.
```
kenjenkins, line 118: I wonder if we should aim to avoid duplication between this page and the main PPL reference page. Would it be better to condense some of this information here? Looks like there is already a link to the main PPL page below.

(Up to you, no strong preference from me.)

content/docs/internals/ppl.mdx
```diff
@@ -123,7 +133,10 @@ Although these policies are equally effective, we recommend using just one opera
 
 ## Criteria
 
-Criteria in PPL are represented as an object where the key is the name and optional sub-path of the criterion, and the value changes depending on which criterion is used. A sub-path is indicated with a `/` in the name:
+In PPL, a criterion defines a specific condition to evaluate, such as a user’s email or device type.
+
+- Each criterion is an object where the key is the criterion name (optionally with a sub-path, delimited by `/`), and the value specifies what to match or compare.
```
kenjenkins, line 138: I wonder if we should expand on this even a bit more? Let me know what you think of this:

> - Each criterion is represented by an object with a single key/value pair, where the key is the criterion type.
> - The format of the value varies depending on the criterion type.
> - For some criteria the key accepts a sub-path, delimited by `/`. For example: `claim/family_name`.
> - Some criteria do not use their value. For example: `accept`, `reject`, and `authenticated_user`. In this case, the value can be anything.

alexandreLamarre:
I think we're good to merge

PR #2006, opened 2025-09-29
ppl: clarify mcp_tool usage

wasaga:
Ref: https://linear.app/pomerium/issue/ENG-2960/pplmcp-tool-only-return-true-for-the-underlying-matcher


nickytonline (approved):
🚢

PR #2009, opened 2025-10-01
revert broken SVG images

kenjenkins:
It looks like some of the SVG text size, color, and placement were unintentionally changed in https://github.com/pomerium/documentation/pull/1726, and some images now have the additional text "Viewer does not support full SVG 1.1". Revert these images to their previous state.


rjbeers (approved):
All fixed! Thanks

PR #2011, opened 2025-10-02
Add a list of prerequisites before building Core from source

agocs:
Listing some tools I needed to install before I could build Pomerium Core. I debated mentioning `brew` but decided against, as that is Mac specific. The rest of these tools should apply to any (reasonable) OS.

### DIFF
--- a/content/docs/deploy/core.mdx
+++ b/content/docs/deploy/core.mdx
@@ -121,6 +121,17 @@ If you plan to run on port 443 in a rootless environment, you may need extra [ca
 
 If you prefer building from source:
 
+### Prerequisites
+
+In order to run `make`, you need the following:
+
+1. [NPM](https://github.com/npm/cli)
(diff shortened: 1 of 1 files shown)


content/docs/deploy/core.mdx
```diff
@@ -121,6 +121,17 @@ If you plan to run on port 443 in a rootless environment, you may need extra [ca
 
 If you prefer building from source:
 
+### Prerequisites
+
+In order to run `make`, you need the following:
+
+1. [NPM](https://github.com/npm/cli)
+2. [Yarn](https://classic.yarnpkg.com/en/docs/install)
+3. [Docker](https://www.docker.com/) (optional)
```
kenjenkins, line 130: Might be nice to mention what you would need docker for — maybe something like this?

```suggestion
3. [Docker](https://www.docker.com/) (optional, used in some tests)
```

kenjenkins (commented):
Thanks for the update!

PR #2021, opened 2025-10-16
ENG-2991: Prepare for 0.31 branching

rjbeers:
- Bump all version references to 0.31 or 0.31.0.
- Update the robot.txt for scaping the latest docs
- Update the version drop down and version list for 0.31


rjbeers:
This should be everything ahead of creating a 0-31-0 branch and moving traffic to it.
Follow up PR would then be to switch things to pre-release and another to turn off bot scraping for 0.30.

PR #2043, opened 2025-12-11
documentation: update docs for k8s per-route identity provider settings

calebdoxsey:
Update documentation for https://github.com/pomerium/ingress-controller/pull/1283


kenjenkins (approved):
Should we also add the new annotation to the list on the https://www.pomerium.com/docs/deploy/k8s/ingress page? I can do that separately if you like.

PR #1896, opened 2025-06-27
docs: add tabbed instructions for Core & Zero on quickstart page

desimone:
## Summary
- Added tabbed interface to quickstart page separating Pomerium Zero and Core instructions
- Enhanced user experience by providing clear deployment path options

## Changes
- Changed page title from "Pomerium Zero Quickstart" to "Pomerium Quickstart"
- Added Tabs component with separate content for Zero and Core
- Added comprehensive Core setup instructions including configuration, secrets generation, and identity provider setup
- Updated page description to cover both deployment options

Fixes #1737

Issue #1737, opened 2025-01-20
Add tabbed instructions for Core & Zero on "Get Started"


desimone:
**Add tabbed instructions for Core & Zero on "Get Started"**  
_Follow-up to [#1726](https://github.com/pomerium/documentation/pull/1726)_

What's incorrect or missing?  
- No direct path for users to choose Pomerium Core or Zero instructions  

What's the resolution?  
- Introduce tabs or toggles on the "Get Started" page for each offering  
- Make sure the differences in steps are clearly separated  

Follow-up question  
- Are the steps for Core vs. Zero significantly different, or mostly similar with a few key changes?

PR #2059, opened 2026-01-06
docs: rewrite Core quickstart for real domains with autocert

desimone:
## Summary

Rewrites the Pomerium Core quickstart for a better first-run experience by assuming users have a real, publicly accessible domain with autocert.

Addresses [Ken's review feedback](https://github.com/pomerium/documentation/pull/1896#pullrequestreview-3623682742) on #1896.

Relates to #1737

## Changes

- **Zero promo**: Tip directing users without prerequisites to Pomerium Zero
- **Deployment methods note**: Mentions binaries/packages as alternatives to Docker
- **Better example domain**: Uses `*.pomerium.yourdomain.com` pattern
- **Secrets explained**: Documents what shared_secret and cookie_secret do, with reference links
- **IdP-agnostic**: Links to identity provider docs instead of Google-specific setup
- **Autocert**: Uses `autocert: true` for automatic Let's Encrypt certificates
- **Section reorder**: Generate secrets → IdP setup → Config → Docker Compose
- **Docker Compose cleanup**: Removed obsolete config, added cert cache volume

### DIFF
--- a/content/docs/get-started/quickstart.mdx
+++ b/content/docs/get-started/quickstart.mdx
@@ -272,7 +272,7 @@ services:
     image: pomerium/pomerium:latest
     volumes:
       - ./config.yaml:/pomerium/config.yaml:ro
-      - pomerium-certs:/data/autocert
+      - pomerium-cache:/data/
     ports:
       - '443:443'
       - '80:80'
@@ -281,10 +281,10 @@ services:
     image: pomerium/verify:latest
 
 volumes:
-  pomerium-certs:
+  pomerium-cache:
 ```
 
-The `pomerium-certs` volume stores Let's Encrypt certificates so they persist across container restarts.
+The `pomerium-cache` volume will store Let's Encrypt certificates so they persist across container restarts.
 
 ## Deploy Pomerium Core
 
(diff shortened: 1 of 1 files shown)


content/docs/get-started/quickstart.mdx
```diff
@@ -276,22 +245,11 @@ Create a `config.yaml` file with your Pomerium configuration:
...
-# Identity provider settings
-idp_provider: REPLACE_WITH_IDP_PROVIDER
-idp_client_id: REPLACE_WITH_IDP_CLIENT_ID
-idp_client_secret: REPLACE_WITH_IDP_CLIENT_SECRET
-
 # Routes
 routes:
   - from: https://verify.pomerium.yourdomain.com
     to: http://verify:8000
+    pass_identity_headers: true
```
desimone: good catch ty.

content/docs/get-started/quickstart.mdx
```diff
@@ -276,22 +245,11 @@ Create a `config.yaml` file with your Pomerium configuration:
 # Pomerium will automatically obtain certificates from Let's Encrypt
 autocert: true
 
-# Authenticate service URL (must match your DNS)
```
desimone: I agree with the change to use hosted authenticate here. Maybe we keep a comment letting folks know they can use a custom / self hosted option later.

content/docs/get-started/quickstart.mdx
```diff
@@ -239,35 +237,6 @@ Your project will contain:
...
-
-Pomerium requires two secrets:
-
-- [**Shared Secret**](/docs/reference/shared-secret): Used to authenticate requests between Pomerium services
-- [**Cookie Secret**](/docs/reference/cookies#cookie-secret): Used to encrypt and sign session cookies
-
-Generate both secrets:
-
-```bash
-# Generate shared secret
```
desimone: I could go either way here and less is probably more in a quick start so I'm biased towards your change of removing this but... my only concern with removing the secrets here is that -- and my memory is hazy here -- I recall this causing confusion for people who churned environments quickly getting frustrated with SSO behavior. It was a minority of fairly non technical users but still something I have scar tissue about. I think PZ is now the better option for that cohort but wanted to put my rationale down.

kenjenkins (approved):

PR #2118, opened 2026-03-18
docs: add visual qa smoke script

desimone:
Add a repeatable visual QA smoke script for the docs site.

It checks a curated set of fragile pages in desktop and mobile, and it can also walk all current `/docs` routes while recording the two pre-existing Kubernetes React recoverable warnings as warnings instead of hard failures.

Closes https://linear.app/pomerium/issue/ENG-3767/docs-visual-qa-smoke-script


package.json
```diff
@@ -16,7 +16,9 @@
     "format": "prettier --write .",
     "format-check": "prettier --check .",
     "precommit": "npm run format-check && npm run cspell",
-    "cspell": "cspell \"**/*\""
+    "cspell": "cspell \"**/*\"",
+    "visual-qa": "node scripts/visual-qa.mjs",
+    "visual-qa:all-docs": "node scripts/visual-qa.mjs --all-docs --desktop-only --skip-screenshots"
```
nickytonline: are we going to run this as part of CI or is this really only meant as a local gut check before pushing?

PR #2119, opened 2026-03-18
docs: normalize repo formatting with prettier

desimone:
Normalize the docs repo with Prettier, fix the config resolution bug that was causing drift, and add the main normalization commit to `.git-blame-ignore-revs`.

This is still primarily a formatting PR, but it also fixes the reason the formatting kept disagreeing across editor, CLI, and CI.

## Root cause

- Prettier was resolving config from `package.json`, not `.prettierrc`, so the repo's `singleQuote: true` setting was not actually being applied.
- The repo also had stale `pre-commit` framework wiring copied from elsewhere, including a script that pointed at a nonexistent `ui/` directory.

## What changed

- `.prettierrc` is now the single source of truth for Prettier in this repo.
- The existing `singleQuote: true` behavior is preserved.
- The CI `pre-commit` check remains, but it now runs the repo scripts directly: `yarn format-check` and `yarn cspell`.
- The Python `pre-commit` framework wiring was removed from this repo (`.pre-commit-config.yaml` and the stale helper script).
- A repo-root `yarn check` command now runs the same formatting and spelling checks locally.
- The repo-wide formatting pass was rerun under the corrected config.

## Subtle diffs reviewers should expect

- Markdown tables are often re-aligned and padded. These should be presentation-only diffs, not content changes.
- JS/TS/MDX imports may be reordered by `@trivago/prettier-plugin-sort-imports`.
- Quote style in JS/TS/MDX is now consistently single quotes again.
- JSON files, and some JSON-looking examples, may have key order changes because `prettier-plugin-sort-json` is still enabled.
- Long relative imports such as `../../../../src/components/EnterpriseConfiguration` are pre-existing in the docs and are not being introduced as a new pattern by this PR.

## Verification

- `yarn format-check`
- `yarn cspell`
- `yarn build`

`yarn build` still emits the pre-existing broken-anchor warnings and the MUI X license warning, but completes successfully.

Closes https://linear.app/pomerium/issue/ENG-3765/repo-wide-prettier-normalization-for-documentation

### DIFF
--- a/content/docs/reference/x-forwarded-for-settings.mdx
+++ b/content/docs/reference/x-forwarded-for-settings.mdx
@@ -156,11 +156,11 @@ Example access log entry:
   "forwarded-for": "192.0.5.1,10.1.2.3",
   "ip": "10.1.2.3",
   "level": "info",
-  "message": "http-request",
   "method": "GET",
   "path": "/",
   "service": "envoy",
-  "time": "2024-07-08T16:42:48-07:00"
+  "time": "2024-07-08T16:42:48-07:00",
+  "message": "http-request"
 }
 ```
 
@@ -179,11 +179,11 @@ Example access log entry:
   // highlight-next-line
   "ip": "192.0.5.1",
   "level": "info",
-  "message": "http-request",
   "method": "GET",
   "path": "/",
   "service": "envoy",
-  "time": "2024-07-08T16:42:48-07:00"
+  "time": "2024-07-08T16:42:48-07:00",
+  "message": "http-request"
 }
 ```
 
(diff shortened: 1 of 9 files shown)


content/docs/reference/x-forwarded-for-settings.mdx
```diff
@@ -153,14 +153,14 @@ Example access log entry:
 
 ```json
```
kenjenkins: I think maybe we should avoid auto-formatting our actual docs content? The log fields in this example output are ordered a certain way, so re-ordering them makes this example less true to life.

PR #2196, opened 2026-05-14
docs: add SSH over TCP/443 L4 edge guide

desimone:
# docs: add SSH over port 443 L4 edge guide

## Summary

Adds a focused guide for SSH TCP routes when users can only egress on port 443 and Pomerium is behind an L4 passthrough edge.

The guide clarifies the main source of confusion: the port in `from: tcp+https://HOST:PORT` is the route-match port carried in the HTTP CONNECT authority. It is not the user's outbound firewall port or the Pomerium listener port.

## What changed

- Added `docs/guides/ssh-tcp-l4-passthrough`.
- Added a runnable Docker Compose example under `content/examples/ssh-tcp-l4-passthrough` with NGINX `stream`, Pomerium, an SSH target, and an in-network client with `pomerium-cli`.
- Linked the guide from the TCP and tunneled SSH docs.
- Positioned Native SSH Access as the preferred SSH experience for most deployments, with this guide scoped to TCP tunneling cases.
- Added production notes for L4/WAF behavior, `--pomerium-url`, SSH idle timeouts, PROXY protocol, and `route_not_found` caused by route-match port mismatches.

## Validation

- Ran the example end to end on OrbStack:
  - Generated demo certificates with `./gen-certs.sh`.
  - Started the stack with `docker compose up -d --build`.
  - Confirmed `pomerium-cli` connected through the NGINX L4 edge on port 443.
  - Confirmed SSH through the tunnel returned `sshd`, `demo`, and `Linux 6.19.13-orbstack-gbd1dc07b8cf4 aarch64`.
  - Confirmed Pomerium authorize logs matched `ssh.localhost.pomerium.io:22` with `allow=true`.
  - Confirmed Pomerium access logs returned `response-code=200` for CONNECT.
  - Confirmed NGINX stream logs showed bytes in both directions.
- Ran `./content/examples/ssh-tcp-l4-passthrough/gen-certs.sh` from the repo root to verify the script is cwd-safe.
- Ran `docker compose config`.
- Ran `yarn format`.
- Ran `yarn format-check`.
- Ran `yarn cspell "**/*"`.
- Ran `yarn build`.

`yarn build` passes and still reports existing unrelated broken-anchor warnings in the enterprise configure and MCP reference docs.

## AI assistance

Codex drafted and edited the guide, implemented Ken's review feedback, and ran the local verification. Claude provided advisory review; I adopted the useful follow-up items for script cwd handling, policy clarity, Docker DNS alias wording, and cleanup consistency. I did not adopt pinned image tags because this example intentionally follows current releases.


content/docs/capabilities/non-http/examples/ssh.mdx
```diff
@@ -30,6 +30,10 @@ By tunneling SSH connections through your Pomerium service:
 
 :::tip
 This example assumes you've already [created a TCP route](/docs/capabilities/non-http/tcp#configure-tcp-routes) for this service.
+:::
+
+:::tip TCP/443-restricted networks
+If clients can only reach outbound TCP/443 or Pomerium sits behind an L4 edge, see [SSH over TCP/443 through an L4 edge](/docs/guides/ssh-tcp-l4-passthrough).
```
kenjenkins: nit: "TCP/443" looks odd to me. I don't think I've seen this notation before.

Also SSH itself uses TCP, right? Is the distinction here really HTTP vs TCP?

I might suggest just "port 443" here instead (and throughout the guide).

content/docs/capabilities/non-http/tcp.mdx
```diff
@@ -46,6 +46,17 @@ Otherwise, the HTTP proxy in front of Pomerium must know how to properly handle
 
 :::
 
+## Two ports, one transport
+
+A TCP route involves two ports that are easy to confuse:
+
+1. The **transport port** the client uses to reach Pomerium. If no explicit port is set in `--pomerium-url`, `pomerium-cli` opens its HTTPS connection on TCP/443. You do not configure that on the route; it's whichever port Pomerium's [`address`](/docs/reference/address), or the edge in front of it, is exposed on.
```
kenjenkins: The wording here is complicated. Would this be a little clearer?

```suggestion
1. The **transport port** the client uses to reach Pomerium. This should match the port in Pomerium's [`address`](/docs/reference/address) setting (unless you are using another reverse proxy in front of Pomerium that is listening on a different port). If this is different from the default https port 443, you'll need to pass the `--pomerium-url` option and include the port number in the URL.
```

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+For most SSH deployments, [Native SSH Access](/docs/capabilities/native-ssh-access) provides a better user experience with OAuth login, ephemeral certificates, and SSH-aware access controls. Use this TCP tunneling pattern when you need SSH to travel over the same HTTPS/TCP 443 path as other Pomerium traffic, cannot change SSH server configuration, or need the same pattern for other TCP protocols.
+
+The key configuration detail is that the port in `from: tcp+https://HOST:PORT` is the **route-match port** carried in the HTTP CONNECT authority. It is not the user's network egress port. For SSH, use `:22` in the TCP route and let `pomerium-cli` reach Pomerium over HTTPS/TCP 443.
+
+## Recommended architecture
+
+```mermaid
+flowchart LR
+  C[SSH client<br/>pomerium-cli] -->|HTTPS CONNECT<br/>TCP/443| E[L4 edge<br/>TCP passthrough]
+  E -->|same TLS stream<br/>TCP/443| P[Pomerium<br/>TLS terminates here]
```
kenjenkins: nit: "TLS terminates" → "terminates TLS" ?

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+The `and` block is shown so you can add group, device, MFA, or claim requirements under the same policy. See [Pomerium Policy Language](/docs/internals/ppl) for full policy syntax.
+
+Use the same destination in SSH:
+
+```ssh-config
+Host ssh-prod
+  HostName ssh.example.com
+  Port 22
+  User testuser
+  ProxyCommand pomerium-cli tcp --listen - %h:%p --pomerium-url https://pomerium.example.com
```
kenjenkins: The `--pomerium-url` option is not needed here,  because we're using the default https port 443, right?

static/examples/ssh-tcp-l4-passthrough/README.md
kenjenkins, line 1: It looks like the other examples in this repo are in content/examples. Should this one go there as well?

If the guide instructs to clone to repo, then we shouldn't need to host these files on the docs site, do we?

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+
+You might also see an unrelated `GET` from `pomerium-cli` with `"response-code":404`. Ignore that entry; the successful proof is the `CONNECT` log.
+
+For log field definitions, see [Authorize Log Fields](/docs/reference/authorize-log-fields) and [Access Log Fields](/docs/reference/access-log-fields).
+
+## Fronting proxy example
+
+For NGINX, use `stream {}` for L4 passthrough:
+
+```nginx
```
kenjenkins: To avoid duplicating these code snippets between this page and the examples folder, we could instead import the nginx.conf file and reference it here.

There's an example of this technique in the Enterprise install page: https://github.com/pomerium/documentation/blob/e81fcea5c8f3cf6069dc293c4fd46c2657687b48/content/docs/deploy/enterprise/install.mdx?plain=1#L191-L194

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+stream {
+    server {
+        listen 443;
+        proxy_pass pomerium:443;
+        proxy_protocol on;
+    }
+}
+```
+
+Then enable Pomerium's bootstrap [`use_proxy_protocol`](/docs/reference/use-proxy-protocol) setting:
```
kenjenkins: I think mentioning "bootstrap" here would only make this more confusing?

```suggestion
Then enable Pomerium's [`use_proxy_protocol`](/docs/reference/use-proxy-protocol) setting:
```

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+
+```yaml
+use_proxy_protocol: true
+```
+
+Only enable this when every connection to the Pomerium listener comes from a PROXY-protocol-speaking edge. Direct clients that do not send the PROXY header will fail.
+
+## Why `:443` in `from` does not work for SSH
```
kenjenkins: I believe port 443 could also work for ssh. You would need to pass it to the `ssh` command using `-p 443`.

Let me try out this example but with port 443 instead....

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+
+   ```bash
+   git clone https://github.com/pomerium/documentation
+   cd documentation/static/examples/ssh-tcp-l4-passthrough
+   ```
+
+1. Generate a demo CA and server certificate.
+
+   ```bash
+   mkdir -p certs
```
kenjenkins: This isn't needed because the `certs` directory is created when cloning the repo.

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+     -config certs/openssl.cnf
+
+   openssl x509 -req -days 365 \
+     -in certs/pomerium.csr \
+     -CA certs/ca.crt \
+     -CAkey certs/ca.key \
+     -CAcreateserial \
+     -out certs/pomerium.crt \
+     -extensions v3_req \
+     -extfile certs/openssl.cnf
```
kenjenkins: This is quite a lot! Would it make sense to move this into a `gen-certs.sh` script in the example directory?

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+1. Review the demo TCP route in `config/pomerium.yaml`.
+
+   ```yaml
+   routes:
+     - from: tcp+https://ssh.localhost.pomerium.io:22
+       to: tcp://sshd:2222
+       allow_public_unauthenticated_access: true
+   ```
+
+   The route is public for the local example. If you remove `allow_public_unauthenticated_access`, configure a real IdP and policy before testing authenticated access.
```
kenjenkins: This last sentence feels a little backwards to me?

Maybe: "The route is public in this example because we're mainly interested in demonstrating connectivity over port 443. In production you'll want to configure a real IdP and policy rules."

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+        sleep 0.2
+      done
+      cat /tmp/pomerium-cli.log
+      exit 1'
+   ```
+
+1. Run an SSH command through the tunnel.
+
+   ```bash
+   docker compose exec -T client sh -lc \
```
kenjenkins: I think all these `-T` options will not be needed if a human is following this example.

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+   The route is public for the local example. If you remove `allow_public_unauthenticated_access`, configure a real IdP and policy before testing authenticated access.
+
+1. Start the stack.
+
+   ```bash
+   docker compose up -d --build
+   docker compose ps
+   ```
+
+1. Confirm that the client resolves the SSH hostname to the L4 edge and can reach it on TCP/443.
```
kenjenkins: I might suggest skipping this step? I'm not sure what it demonstrates.

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+       --browser-cmd /bin/true \
+       --listen 127.0.0.1:2222 \
+       >/tmp/pomerium-cli.log 2>&1'
+   ```
+
+   `--browser-cmd /bin/true` suppresses browser launch in the headless client container. Do not copy it into production commands that need interactive login. Production SSH `ProxyCommand` configuration usually uses `--listen -` instead.
+
+   In this local stack, `pomerium-cli` dials `ssh.localhost.pomerium.io:443`, and Docker resolves that hostname to the `nginx` L4 edge.
+
+1. Wait for the listener.
```
kenjenkins: I'm not sure we really need this step either? If a human is following this example I doubt they'd be able to run these steps quickly enough where the startup latency is going to cause a problem?

content/docs/guides/ssh-tcp-l4-passthrough.mdx
```diff
@@ -0,0 +1,324 @@
...
+      exit 1'
+   ```
+
+1. Run an SSH command through the tunnel.
+
+   ```bash
+   docker compose exec -T client sh -lc \
+     "sshpass -p demo-password ssh \
+       -o StrictHostKeyChecking=no \
+       -o UserKnownHostsFile=/dev/null \
```
kenjenkins: We don't need this option if we already have `-o StrictHostKeyChecking=no` do we?

content/docs/guides/ssh-tcp-l4-passthrough.mdx
kenjenkins: Yes, changing the from route to use `:443` and the `pomerium-cli` command to use `ssh.localhost.pomerium.io:443` seems to work fine for me.

desimone:
thx ken, pushed a pass at this in 6e1fc761.

main changes: moved the example to content/examples, swapped the long cert block for gen-certs.sh, imported nginx.conf, cleaned up the human commands, and reworked the :443 bit so it says the real thing: the route port has to match the connect authority. normal ssh stays :22 while the outer transport is still https on port 443.

also reran the compose proof from the moved path and checked ssh output + pomerium/nginx logs.

kenjenkins (approved):

PR #2220, opened 2026-06-02
docs: add Forgejo reverse-proxy SSO guide with runnable example

desimone:
Pomerium has guides for plenty of self-hosted apps but nothing for Forgejo or Gitea, even though a Pomerium-gated Git server is a common setup (we run one ourselves as a private mirror). This adds that guide: how to put Forgejo behind Pomerium and use reverse-proxy authentication so people sign in once with their existing identity and Forgejo provisions their account from the headers Pomerium forwards, with no second password. The runnable Docker Compose example lives in `content/examples/forgejo/` and is imported into the guide, the same way the guacamole and tiddlywiki guides do it.

The section worth reading closely is the security model. Forgejo's reverse-proxy auth trusts the identity header from any peer that can reach it, and `REVERSE_PROXY_TRUSTED_PROXIES` does not change that; it only affects how Forgejo reads `X-Forwarded-For`. So network isolation is the real trust boundary, and the guide says that plainly instead of implying the trusted-proxies setting is what protects you. I checked this against Forgejo 15.0.2 before writing it up: a spoofed header from an untrusted container was accepted and even auto-created a user, which is exactly why the guide leans so hard on keeping Forgejo reachable only through Pomerium. The verification, admin-bootstrap, and troubleshooting steps all describe what actually happened on a real instance rather than what should happen in theory.

## AI assistance

Claude (Opus 4.8) drafted the guide and the example files and ran the validation (Docusaurus build, cspell, prettier, and a live Forgejo e2e). I reviewed it and signed off. The one correction worth calling out: the e2e disproved an earlier draft's claim that the spoof test would be rejected, so the security framing was rewritten to match what Forgejo actually does. A second review pass with Codex caught a leftover contradiction in the Compose comment, which is fixed.


content/docs/guides/forgejo.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+# Secure Forgejo with Pomerium
+
+Learn how to add authentication, authorization, and single sign-on to a [Forgejo](https://forgejo.org/) instance with Pomerium. The same reverse-proxy authentication pattern applies to [Gitea](https://about.gitea.com/), which Forgejo is a soft-fork of, but this guide uses the Forgejo container image and `FORGEJO__` environment variables.
+
+## Background
+
+Forgejo is a self-hosted Git service. Like most Git servers, it has its own local username and password database. Without single sign-on, every user has a second set of credentials in addition to your identity provider.
+
+Forgejo also supports [reverse-proxy authentication](https://forgejo.org/docs/latest/admin/setup/reverse-proxy/): it trusts an identity supplied in an HTTP header by a proxy in front of it. Pomerium is exactly that proxy. After Pomerium authenticates a user against your identity provider, it forwards the user's identity as request headers; Forgejo reads those headers, signs the user in, and (optionally) auto-creates the account on first visit. The result is single sign-on for the Forgejo **web UI** with no second password.
```
kenjenkins: nit: the phrase "which Forgejo is a soft-fork of" feels awkward to me.

I have a few other suggestions for tweaking the flow here, feel free to take or leave.


```suggestion
Learn how to add authentication, authorization, and single sign-on to a [Forgejo](https://forgejo.org/) instance using Pomerium. Forgejo is a self-hosted Git service with a web UI.

A similar setup can also be used with [Gitea](https://about.gitea.com/).

## Background

Like many applications, Forgejo has its own local username and password database. Without single sign-on, every user would need to maintain a second set of credentials.

However, Forgejo also supports [reverse-proxy authentication](https://forgejo.org/docs/latest/admin/setup/reverse-proxy/): it trusts an identity supplied in an HTTP header by a proxy in front of it. Pomerium is exactly that proxy. After Pomerium authenticates a user against your identity provider, it forwards the user's identity in request headers. Forgejo reads these headers, signs the user in, and can even automatically create an account on first visit. The result is single sign-on for the Forgejo web UI with no second password.
```

content/docs/guides/forgejo.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+Reverse-proxy authentication means Forgejo believes the identity headers it receives. Forgejo does not verify Pomerium's signed `X-Pomerium-Jwt-Assertion` for login, so this integration depends entirely on **network isolation**: Forgejo must be reachable **only** through Pomerium. `REVERSE_PROXY_TRUSTED_PROXIES` does not gate this; it only controls how Forgejo reads `X-Forwarded-For`. Reverse-proxy auth trusts the identity header from any peer that can reach Forgejo, so the network path is the entire boundary.
+
+If another client can connect to Forgejo directly, it can set the same identity headers and impersonate a user. In Docker Compose, do not publish Forgejo's HTTP port and keep only Pomerium on its network. In Kubernetes, a `ClusterIP` service alone is not enough if arbitrary pods can connect to Forgejo; use a policy-enforcing CNI (container network interface) and a `NetworkPolicy`, or another network control, to restrict the backend to Pomerium.
+
+When an upstream app can verify the signed Pomerium JWT directly, prefer that model. Forgejo does not support that login mode today, so this guide uses header-based reverse-proxy authentication with a narrow trust boundary.
```
kenjenkins: I'd suggest trimming this a bit.


```suggestion
Reverse-proxy authentication means Forgejo believes the identity headers it receives. Forgejo does not verify Pomerium's signed `X-Pomerium-Jwt-Assertion` for login, so this integration depends entirely on **network isolation**: Forgejo must be reachable **only** through Pomerium. If another client can connect to Forgejo directly, it can set the same identity headers and impersonate a user.

In Docker Compose, do not publish Forgejo's HTTP port and keep only Pomerium on its network. In Kubernetes, a `ClusterIP` service alone is not enough if arbitrary pods can connect to Forgejo; use a policy-enforcing CNI (container network interface) and a `NetworkPolicy`, or another network control, to restrict the backend to Pomerium.

Whenever possible, prefer using Pomerium JWT authentication, which does not depend on network isolation. Forgejo does not support that login mode today, so this guide uses header-based reverse-proxy authentication.
```

content/examples/forgejo/forgejo.pomerium-config.md
```diff
@@ -0,0 +1,30 @@
...
+    to: http://forgejo:3000
+    pass_identity_headers: true
+    # Defense in depth: strip any client-supplied copies of these headers on the
+    # way in, so the upstream only ever sees Pomerium-injected values.
+    remove_request_headers:
+      - X-Pomerium-Jwt-Assertion
+      - X-Pomerium-Jwt-Assertion-For
+      - X-Pomerium-Sub
+      - X-Pomerium-Claim-Email
+      - X-Pomerium-Claim-Name
```
kenjenkins: To me, this feels more like sowing doubt than defense in depth. Do we really need this?

content/examples/forgejo/forgejo.docker-compose.md
```diff
@@ -0,0 +1,62 @@
...
+
+  forgejo:
+    image: codeberg.org/forgejo/forgejo:15.0.2
+    # Do not publish Forgejo's port to the host. Reverse-proxy auth trusts the
+    # identity header from ANY peer that can reach forgejo:3000, and
+    # REVERSE_PROXY_TRUSTED_PROXIES does NOT change that (it only affects
+    # X-Forwarded-For parsing). Network isolation is the trust boundary: keep
+    # only Pomerium on this network and attach no untrusted containers.
+    expose:
+      - '3000'
```
kenjenkins: `expose` is generally not needed — any other container on the same network can access any port already.

content/docs/guides/forgejo.mdx
```diff
@@ -0,0 +1,161 @@
...
+## Troubleshooting
+
+- **Pomerium returns `403`.** The route policy still allows only `user@example.com`. Replace it with your user or domain.
+- **You reach Forgejo but are not signed in or auto-provisioned.** The header names in `config.yaml` (`jwt_claims_headers`) and in Forgejo (`REVERSE_PROXY_AUTHENTICATION_USER`/`_EMAIL`/`_FULL_NAME`) must match exactly, and `pass_identity_headers` must be set on the route.
+- **Sign-in fails on the username.** Forgejo rejects usernames containing characters such as `@`. Use a stable, already-valid claim for the username (this guide uses `sub`).
+- **A direct request to the backend signs in.** Expected: reverse-proxy auth trusts the header from any peer. It is why Forgejo must stay reachable only through Pomerium.
+
+## Clean up
+
+Stop the stack with `docker compose down`. Add `-v` only to also delete the `forgejo-data` volume and all Forgejo state. To undo the integration while keeping data, remove the Forgejo route from `config.yaml` and unset the `ENABLE_REVERSE_PROXY_*` settings, after creating a break-glass admin as above.
```
kenjenkins: nit: "only to also delete" feels awkward to me, maybe something like this instead?

```suggestion
Stop the stack with `docker compose down`. To delete the `forgejo-data` volume (with all Forgejo state), use `docker compose down -v`. To undo the integration while keeping data, remove the Forgejo route from `config.yaml` and unset the `ENABLE_REVERSE_PROXY_*` settings, after creating a break-glass admin as above.
```

kenjenkins (approved):
Looks good, left a few suggestions.

nickytonline (commented):
We have a skill/rubric for guides that I ran against this guide. Overall, solid. Ran the skill in Claude with Opus 4.8

Some minor tweaks:

  **Total Weighted Score: 42 / 45** (score band: Publish-ready, 38–45)

  **Publish blockers present: Yes (acronym rule)**

  The score lands in the publish-ready band, but the skill's hard rules override
  score. Two acronym-expansion violations:

  1. LTS is never expanded. Line 96: "The example pins Forgejo to a current 
  LTS". LTS is not on the exempt internet-infrastructure list. Fix: "pins 
  Forgejo to a current Long-Term Support (LTS) release".
  2. CNI is defined in reverse order. Lines 54 and 129: "a policy-enforcing CNI 
  (container network interface)". The rule requires full term first, acronym in
  parens. Fix: "a policy-enforcing container network interface (CNI)" on first
  use (line 54); line 129 can then stay CNI.

  No em dashes, en dashes, or emojis found. CIDR and SSH I'm treating as exempt
  under the "widely-understood internet infrastructure" clause.

  Recommendations, ordered by impact

  1. Fix the two acronyms (blocker, ~2 minutes). This is all that stands between
  the guide and publish-ready.
  2. Add an explicit "Who this is for" line near the intro (e.g. self-hosted
  Forgejo/Gitea operators already running Pomerium). Moves Audience + Outcome
  from 2 → 3 (+1).
  3. Promote Troubleshooting to the recommended table (Symptom | Likely Cause | 
  What to Check | Fix) and add a "where to check next" cue per row (e.g.
  Pomerium logs, docker compose logs forgejo). Moves Troubleshooting 2 → 3 (+2).
  4. Optional: add an Excalidraw request-flow diagram to the architecture
  section. Already a 3, so no score change, but the skill prefers a visual for
  request-flow orientation.

  Publish recommendation: Needs edits (minor)

  Content quality is publish-ready (42/45), but the acronym hard-fail must be
  cleared first. Fixing items 1–3 would bring it to a clean 45/45.

--

⏺ Scan results: no em dashes, en dashes, or emojis anywhere. Two acronym-rule
  issues, both trivial fixes. Here's the full worksheet in the skill's required
  format.

  Guide review: Secure Forgejo with Pomerium

  Reviewed: content/docs/guides/forgejo.mdx plus its two embedded snippets
  (forgejo.docker-compose.md, forgejo.pomerium-config.md).

  Dimension: Audience + Outcome
  Raw Score (0-3): 2
  Weight: 1
  Weighted Score: 2
  Notes: Outcome is excellent: concrete, testable, and explicitly scoped (the
    :::caution states what it does not do — API/Git/SSH stay token-based). But
    audience is only implied; there is no explicit "Who this is for." That gap
    caps it at Good.
  ────────────────────────────────────────
  Dimension: Use-Case Fit + Decision Support
  Raw Score (0-3): 3
  Weight: 1
  Weighted Score: 3
  Notes: States when it fits, the central tradeoff (header auth depends on
    network isolation; prefer JWT but Forgejo can't yet), the scope limit, and a

    real use case (warm-standby mirror).
  ────────────────────────────────────────
  Dimension: Prerequisites + Environment Assumptions
  Raw Score (0-3): 3
  Weight: 2
  Weighted Score: 6
  Notes: Self-hosted/Core guide. Explicitly states a working Pomerium deployment

    is required and links the Quickstart (satisfies the baseline prereq via the
    Core substitution). Lists Docker/Compose + config files, ordered, plus the
    Zero/Enterprise variant. Forgejo image is version-pinned.
  ────────────────────────────────────────
  Dimension: Architecture + Request Flow Context
  Raw Score (0-3): 3
  Weight: 1
  Weighted Score: 3
  Notes: "How you will secure Forgejo" (two halves) + Background give a precise
    narrative of request flow, trust boundaries, and control points. No diagram,

    but the narrative qualifies.
  ────────────────────────────────────────
  Dimension: Procedural Clarity
  Raw Score (0-3): 3
  Weight: 2
  Weighted Score: 6
  Notes: Precise, ordered, copy-pasteable; every non-obvious setting is
    annotated; the placeholder-replacement step is called out explicitly.
  ────────────────────────────────────────
  Dimension: Verification + Checkpoints
  Raw Score (0-3): 3
  Weight: 2
  Weighted Score: 6
  Notes: Browser sign-in (with screenshot + exact expected state), host-port
  curl
    with expected result, docker network inspect, plus the K8s equivalent. Exact

    expected outcomes throughout.
  ────────────────────────────────────────
  Dimension: Troubleshooting + Failure Modes
  Raw Score (0-3): 2
  Weight: 2
  Weighted Score: 4
  Notes: Four realistic entries mapping symptom → cause → fix (solid Good). Held

    below Excellent: no "where to inspect next" pointer per entry and not in the

    recommended table format.
  ────────────────────────────────────────
  Dimension: Security + Risk Posture
  Raw Score (0-3): 3
  Weight: 2
  Weighted Score: 6
  Notes: Strongest dimension. Documents the core risk (header trust), names
    network isolation as the trust boundary, adds defense-in-depth header
    stripping, break-glass admin, and production caveats (pin images).
    Insecure-by-default is actively prevented.
  ────────────────────────────────────────
  Dimension: Operations + Lifecycle Management
  Raw Score (0-3): 3
  Weight: 1
  Weighted Score: 3
  Notes: Applicable (self-hosted, pins images, owns the data volume + admin
    lifecycle). Covers rollback (remove route, unset env), cleanup (down  -v),
    break-glass recovery, and backups (dump, pull-mirror scope caveats).
  ────────────────────────────────────────
  Dimension: Maintainability + Scanability
  Raw Score (0-3): 3
  Weight: 1
  Weighted Score: 3
  Notes: Clear headings, titled fenced blocks, admonitions, consistent naming,
    single-sourced imported snippets. Highly skimmable. (See hard-fail note
  below
     — the structural quality is a 3, but the acronym rule is violated.)

content/examples/forgejo/forgejo.pomerium-config.md
desimone: Removed. With `pass_identity_headers: true`, Pomerium emits these claim headers on every request and overwrites any client-supplied copy (`OVERWRITE_IF_EXISTS_OR_ADD`), so a pre-set value can't reach Forgejo — the explicit strip was redundant. Agreed it read as sowing doubt.

content/examples/forgejo/forgejo.docker-compose.md
desimone: Removed — you're right, same-network containers reach the port regardless. Kept the comment since it explains why there's no host port mapping.

PR #2232, opened 2026-06-06
docs: standardize HedgeDoc guide with sealed validation

desimone:
## What this does

Standardizes the HedgeDoc guide to the shared guide template and backs it with a sealed end-to-end validation fixture, mirroring the Grafana pilot.

The guide is rewritten with the standard sections (What this guide does, When to use, Prerequisites, Configure Pomerium with Zero + Core tabs, Configure HedgeDoc, Run the stack, Verify, Common failure modes, Security considerations, Operations, Next steps). It drops the old mkcert + AWS Cognito flow and the wildcard-cert section in favor of the standard hosted-authenticate reader config (`autocert: true`). The route keeps `allow_websockets: true` for HedgeDoc's live collaborative editor, and `CMD_DOMAIN` / `CMD_PROTOCOL_USESSL` / `CMD_URL_ADDPORT` are set so HedgeDoc builds correct external URLs behind Pomerium's TLS.

HedgeDoc just needs protecting (it does not consume a Pomerium identity header), so this uses the default access checks from the harness: the route must require authentication, and an allowed user must reach the app after login. The fixture runs HedgeDoc plus its Postgres database behind a Pomerium wired to the in-network Keycloak. Validation images are pinned by digest for reproducibility; the reader-facing files keep `:latest` / version tags. HedgeDoc gets a 90s healthcheck `start_period` for its slow first boot.

## Validation

`scripts/validate-guide-fixtures.sh hedgedoc` (sealed E2E, after digest pinning):

```
Running 2 tests using 1 worker

  ✓  1 [chromium] › access.spec.ts:9:5 › unauthenticated request is redirected to the identity provider (533ms)
  ✓  2 [chromium] › access.spec.ts:14:5 › authenticated user reaches the upstream (490ms)

  2 passed (1.3s)
>> hedgedoc: PASS
```

Docs checks (Node 22.22.2): `yarn format-check` clean, `yarn cspell` 0 issues, `yarn build` SUCCESS with no broken links from this guide (the only broken anchors reported are pre-existing in unrelated pages: configure-metrics, upgrading, mcp-bridge).

Image digests verified to resolve to the tags the reader files ship (`hedgedoc:1.10.8`, `postgres:16-alpine`).

## AI assistance

Claude (Opus 4.8) drafted the rewritten guide and the full validation fixture (reader compose/config + byte-identical `.md` mirrors, `validate/compose.validate.yaml`, `validate/pomerium.validate.yaml`), pinned the validation image digests, and ran the sealed validate script plus all docs checks. I verified the work by running `scripts/validate-guide-fixtures.sh hedgedoc` to green twice (once after digest pinning), confirming `yarn build` / `yarn cspell` / `yarn format-check` pass, checking the digests resolve to their tags, and grepping the diff for path/secret/customer leaks.

Reviewer (Claude advisory pass via review-major-change.sh): adopted reconciling `CMD_ALLOW_ANONYMOUS` (removed from the reader compose so HedgeDoc keeps its own login; the security section now explains the implication), adding an Operations section (teardown vs `down -v`, volume persistence, upgrade note), expanding "IdP" on first use, a Zero shared-network note, and a demo-password warning. Rejected the em-dash "hard-fail" finding (the reference Grafana guide uses em dashes; not a codebase rule) and the failure-modes-as-table suggestion (kept bullets to match the established template).


desimone:

PR #2240, opened 2026-06-06
docs: standardize AdGuard guide with sealed validation

desimone:
## What this does

Standardizes the AdGuard Home guide to the Pomerium guide template and adds a sealed end-to-end validation fixture that proves the full flow locally.

AdGuard Home has no SSO and no user directory; its only built-in access control is a single set of HTTP basic-auth credentials. This guide puts AdGuard behind Pomerium so Pomerium handles single sign-on and authorization, then injects AdGuard's basic-auth credential on every upstream request via `set_request_headers`. AdGuard auto-authenticates from that header, so users sign in once through Pomerium and never see AdGuard's own login prompt. This is the general header-injection pattern for any app whose only auth is a static credential.

## Category: header trust (header injection)

Because the upstream trusts the injected credential, the guide's Security section warns that only Pomerium may reach AdGuard (direct bypass), notes the basic-auth secret lives in Pomerium's config, and reminds readers that every allowed user maps to the one shared AdGuard admin, so the route policy should be scoped.

## Validation

The fixture reuses the shared harness (in-network Keycloak, certs, headless Playwright runner). An init container seeds `AdGuardHome.yaml` with a bcrypt admin user so AdGuard boots past its install wizard and enforces basic auth (AdGuard rewrites its config on startup, so the seed is copied into a writable volume rather than mounted read-only). The Pomerium route injects the matching `Authorization: Basic` header.

`assert.spec.ts` asserts three things:

1. An unauthenticated request is redirected to the identity provider.
2. After SSO login, AdGuard's control API (`/control/status`) responds `200` from the injected credential, returning a valid status payload.
3. A request straight to `adguard:3000` (bypassing Pomerium, carrying no credential) is rejected with `401`, proving the credential is what gates access and demonstrating the direct-bypass trust boundary.

Local run (images pinned by digest):

```
Running 3 tests using 1 worker

  ✓  1 [chromium] › assert.spec.ts:10:5 › unauthenticated request is redirected to the identity provider (593ms)
  ✓  2 [chromium] › assert.spec.ts:15:5 › injected basic auth reaches AdGuard's control API authenticated (743ms)
  ✓  3 [chromium] › assert.spec.ts:38:5 › AdGuard rejects a direct request that skips Pomerium (24ms)

  3 passed (1.7s)
>> ad-guard: PASS
```

Docs checks (Node 22.22.2): `prettier --write` clean, `yarn format-check` clean, `yarn cspell` 0 issues, `yarn build` SUCCESS with no new broken links. Renaming the guide from `.md` to `.mdx` (for tabs) required fixing one stale link in `helm.mdx` that pointed at `/docs/guides/ad-guard.md`.

## AI assistance

Claude (Opus) drafted the rewritten guide, the reader-facing `config.yaml` / `docker-compose.yaml` mirrors, the validation fixture (compose, Pomerium config, seeded `AdGuardHome.yaml`, and the Playwright spec), and this PR body. I verified the behavior empirically before writing the assertions: ran the upstream `adguard/adguardhome` image, completed its install API, and confirmed `/control/status` returns `401` without the basic-auth header and `200` with `Authorization: Basic <base64 of admin:...>`. I ran the sealed fixture to a green `>> ad-guard: PASS` (3 tests) and ran the full docs build to SUCCESS. The workspace cross-model review flagged that the doc claimed a `401`-without-header assertion the spec didn't perform; I added that as a real direct-bypass test (also proving the security model) rather than softening the wording, fixed a reader/wizard ordering forward-reference, turned the example header value into an obvious placeholder, and dropped a stale cspell ignore.


desimone:

PR #2249, opened 2026-06-06
docs: standardize ToolJet guide with sealed validation

desimone:
## What this does

Standardizes the ToolJet guide to the shared guides template and adds a sealed end-to-end validation fixture that proves the full SSO gate locally.

- Rewrites `content/docs/guides/tooljet.mdx` to the standard sections with a Pomerium Zero (default) + Pomerium Core tab set, the hosted authenticate service, and the Keycloak self-host fallback. Drops the old Enterprise-console screenshot walkthrough.
- Migrates the example from `content/examples/tooljet/` into `content/examples/guides/tooljet/` with runnable mirrors (`docker-compose.yaml`, `config.yaml`) and byte-identical fenced `.md` copies that the guide imports. Removes the old example dir and the eight `img/tooljet/` screenshots.
- Adds `validate/` (sealed fixture): `compose.validate.yaml` reuses the shared harness and adds ToolJet CE + PostgreSQL; `pomerium.validate.yaml` adds the ToolJet route.

## Category

HTTP-SIMPLE: Pomerium is the front-door gate; ToolJet keeps its own auth and RBAC. Uses the harness default access spec (unauthenticated request redirects to the IdP; an allowed user reaches ToolJet after login). No custom assertion spec.

## ToolJet specifics

ToolJet validates the incoming `Host` against `TOOLJET_HOST`, so the route uses `preserve_host_header: true` (the external route host already equals `TOOLJET_HOST`). The `tooljet/tooljet-ce` image runs its DB setup automatically on first boot and bundles Redis, so the only external dependency is PostgreSQL; the upstream gets a 240s healthcheck `start_period` and Pomerium/test-runner gate on it being healthy. Requires `TOOLJET_HOST`, `LOCKBOX_MASTER_KEY`, `SECRET_KEY_BASE`, `PG_*`, and `TOOLJET_DB_*`.

## Validation

Sealed fixture is green (re-confirmed after pinning all validation images by digest):

```
>> tooljet: running validation checks
  ✓  1 [chromium] › access.spec.ts:9:5 › unauthenticated request is redirected to the identity provider (634ms)
  ✓  2 [chromium] › access.spec.ts:14:5 › authenticated user reaches the upstream (2.0s)
  2 passed (3.0s)
>> tooljet: PASS
```

Docs checks (Node 22.22.2): `prettier --write` clean, `yarn format-check` clean, `yarn cspell` 0 issues, `yarn build` SUCCESS. The only broken anchors reported by the build are pre-existing and unrelated (enterprise/configure-metrics, deploy/upgrading, guides/mcp-bridge).

## AI assistance

Claude (Opus 4.8) drafted the rewritten guide, the migrated example files, and the sealed validation fixture, and iterated on the ToolJet env/DB configuration. I verified the work by running the sealed `validate-guide-fixtures.sh tooljet` harness to a real green PASS (three times, including after digest pinning and after switching `host_rewrite_header` to the correct `preserve_host_header`), and by running prettier, format-check, cspell, and a full `yarn build` locally.

Reviewer (Claude via review-major-change.sh): adopted the `host_rewrite_header: true` -> `preserve_host_header: true` fix (the former takes a header name, not a boolean, so it was a silent no-op), added a test-only-secrets comment and a note that the harness verifies Pomerium gating rather than ToolJet's internal host check. Rejected/already-handled: postgres was already digest-pinned, the `depends_on` chain already gates the test-runner on ToolJet being healthy, and the stale `examples/tooljet`/`img/tooljet` artifacts were already removed (grep-verified zero remaining references).


desimone:

PR #2250, opened 2026-06-06
docs: standardize TiddlyWiki guide with sealed validation

desimone:
## What this does

Standardizes the TiddlyWiki on Node.js guide to the shared guide template and adds a sealed end-to-end validation fixture so the guide's full flow is proven on every change.

TiddlyWiki's Node.js server has no login of its own. Pomerium forwards the authenticated user's email in an unsigned header (`jwt_claims_headers` -> `X-Pomerium-Claim-Email`, with `pass_identity_headers`), and TiddlyWiki's `authenticated-user-header` trusts that value as the logged-in user. The guide leads with the hosted authenticate service (Zero default + Core tabs), links the Keycloak fallback for self-hosting the IdP, and the Security section spells out the unsigned-header trust boundary: TiddlyWiki must be unreachable except through Pomerium because it blindly trusts the header.

## Validation

This is an identity-consuming (header-auth) guide, so the fixture adds an app-specific assertion on top of the default access check. The sealed fixture (`content/examples/guides/tiddlywiki/validate/`) runs Pomerium against the in-network Keycloak, drives the real SSO login headlessly, and asserts TiddlyWiki's `/status` reflects the forwarded identity.

```
>> tiddlywiki: building and starting stack (POMERIUM_URL=https://tiddlywiki.localhost.pomerium.io)
>> tiddlywiki: running validation checks

Running 2 tests using 1 worker

  ✓  1 [chromium] › assert.spec.ts:9:5 › unauthenticated request is redirected to the identity provider (550ms)
  ✓  2 [chromium] › assert.spec.ts:14:5 › header auth signs the authenticated user into TiddlyWiki (763ms)

  2 passed (1.6s)
>> tiddlywiki: PASS
```

Docs checks (Node 22.22.2): `yarn format-check` clean, `yarn cspell` 0 issues, `yarn build` SUCCESS with no TiddlyWiki broken links. Validation images are pinned by digest; the reader-facing files keep `:latest`. The reader `docker-compose.yaml` makes the init step idempotent so repeated `docker compose up` runs work. Old `content/examples/tiddlywiki/` and the orphaned Enterprise console screenshots were removed.

## AI assistance

Claude (Opus 4.8) drafted the rewritten guide, the reader and validation fixtures, and the Playwright assertion spec. I verified manually: ran `./scripts/validate-guide-fixtures.sh tiddlywiki` to green across multiple runs (output above), confirmed the TiddlyWiki `/status` identity reflection against the image directly, and ran the docs format/spell/build checks locally. A cross-model review pass flagged an unexpanded "IdP" acronym, a non-idempotent reader init, a placeholder-domain mismatch, and a mistargeted "signed JWT" link; all four were adopted.


desimone:

PR #2251, opened 2026-06-06
docs: standardize guide structure and validate runnable examples

desimone:
## Summary

This PR standardizes the guides collection and adds validation coverage for runnable guide examples.

## What changed

- Rewrote guides around a consistent reader-first structure.
- Refreshed stale examples, screenshots, links, and version references.
- Added Docker Compose validation fixtures for runnable guides.
- Added `validate/SKIP` markers for guides that require external infrastructure or manual verification.
- Updated CI so guide prose changes can trigger example validation.

### End-to-end fixtures

- grafana
- jenkins
- forgejo
- guacamole
- ad-guard
- llm
- tiddlywiki
- hedgedoc
- code-server
- gitlab

### Gate and reachability fixtures

- cockpit
- transmission
- tooljet
- openclaw-gateway

### Transport fixture

- ssh-tcp-l4-passthrough

### Requires manual verification

These guides depend on Kubernetes, hardware, cloud services, or managed control planes:

- synology
- argo
- istio
- zero-ssh
- jit
- homelab-pipeline
- mcp-bridge
- cloud-run

### Redirected or retired

- helm
- local-mcp

## Reviewer notes

Please focus on:

1. whether the rewritten guides are clear and technically accurate
2. whether Zero and Core paths are explicit enough where the flows differ
3. whether each validation category matches what is actually tested
4. whether retired pages redirect to the right replacement docs

## Local checks

```bash
yarn format-check
yarn cspell
yarn guide-audit
yarn build
scripts/validate-guide-fixtures.sh --selftest
scripts/validate-guide-fixtures.sh <guide-id>
```

## AI assistance

Claude Opus assisted with initial guide rewrites, fixtures, and harness code. I reviewed and edited the final content and ran the validation checks listed above.

PR #2253, opened 2026-06-08
docs(guides): add Immich, Jellyfin, Paperless-ngx, and servarr guides

desimone:
## What this adds

Four new guides for putting high-traffic self-hosted apps behind Pomerium, all at the
front-door authorization tier: the app keeps its own login, and Pomerium adds centralized single
sign-on, group and domain policy, MFA, and an audit trail in front of it rather than replacing the
app's accounts. These are the four apps from the most-requested set that had no
Pomerium guide: Immich (photos), Jellyfin (media), Paperless-ngx (documents), and the servarr
media-automation stack (Sonarr, Radarr, Prowlarr, SABnzbd).

Each guide follows the standardized structure: dual Pomerium Zero (console) and Pomerium Core
(config.yaml) tabs, a runnable docker-compose example with byte-identical `.yaml.md` mirrors, a
mermaid request-flow diagram, and one hand-authored, Tufte-style SVG. The media guides lead with
the caveat that matters most for this tier: a browser SSO gate covers the web UI, but native and
mobile clients (the Immich app, Jellyfin's TV/Chromecast apps, the *arr API keys) authenticate
with their own tokens and cannot complete that browser flow, so they need a deliberate separate
path or gating the whole host breaks them. The servarr guide is
honest that those apps have no OIDC or header identity at all and cross-links the existing
Transmission guide instead of repeating it.

The media routes also call out the long-transfer side of the problem: Immich uploads/downloads and
Jellyfin direct-play/download responses are plain HTTP traffic, so the examples lift the total
request timeout separately from WebSocket support.

Identity framing is now accurate per app rather than a blanket "front-door gate only." Each guide
ships as a front-door gate (no `pass_identity_headers` set, since it defaults to false), but where an
app can actually consume identity, the guide says so and points at the app's own docs as a next step:
Paperless-ngx supports native trusted-header SSO (`PAPERLESS_ENABLE_HTTP_REMOTE_USER`), and Immich
supports native OIDC. Jellyfin core has no SSO (a community plugin adds OIDC for the web UI only), and
the servarr apps have a single shared admin credential with no per-user identity, so for those the
front-door gate genuinely is the ceiling. Deeper worked SSO sections for Paperless-ngx and Immich are
tracked as follow-ups.

## How it's validated

Every guide ships a sealed end-to-end fixture that runs the real upstream plus an in-network
Keycloak behind Pomerium and drives a headless browser: it proves an unauthenticated request is
redirected to the identity provider, an allowed user passes the gate, and the upstream answers
behind it (Immich `/api/server/ping`, Jellyfin `/System/Info/Public`, Paperless `/accounts/login/`,
and each *arr app's status API). The media and servarr validation fixtures put the upstream on a
network path isolated from the test runner and include a negative control proving it is reachable
only through Pomerium.
`scripts/validate-guide-fixtures.sh` passes for all four, and `yarn build`, Prettier, and cspell
are clean. Images are pinned by digest and kept byte-consistent across the reader compose, its
mirror, and the validation compose.

This PR is based on `docs/guides-standardization` because that branch carries the validation
harness; rebase onto main after the harness PR (#2251) lands.

## AI assistance

Drafted with Claude (Opus). Claude authored the guide prose, the example fixtures and validation
specs, and the SVG/mermaid visuals, then iterated each guide against the sealed harness until it
passed. A human reviewed every guide for accuracy and voice, re-ran the full validation suite, and
deployed all four behind Pomerium in a live cluster to confirm the gate end to end before merge.
No Pomerium features were invented; where a guide was unsure it mirrors the existing grafana,
ad-guard, and gitlab guides.


nickytonline (commented):
I know there's some shared files (mainly sidebar I believe), but can this be broken into 4 PRs, one for each guide? Easier to review, and we'd be able to land/revert each one independently (if we needed to).

desimone:
Sure -- that makes sense. Let's get this clean up PR https://github.com/pomerium/documentation/pull/2251  reviewed first since it has the CI pipeline to make sure we dont break things  of the existing.  then i'll resubmit seperate guides.

desimone:
BTW, each of these I am happy to make a video walkthing through I just need help on the editing.

PR #2255, opened 2026-06-09
docs(guides): add Jellyfin guide

desimone:
Jellyfin is a self-hosted media server for movies, shows, and music. This guide puts its web client behind Pomerium so the browser front door is gated by your existing identity provider with single sign-on, group and domain policy, and MFA, while Jellyfin keeps its own accounts underneath. As with the other apps in this set, Pomerium sits in front as a verified-identity gate and audit trail rather than replacing Jellyfin's login.

The catch with Jellyfin is its clients. The native TV apps, Chromecast, and mobile players authenticate straight to the Jellyfin API with Jellyfin tokens and never run the browser SSO redirect, so a route that requires SSO blocks casting and the living-room apps. The guide states that constraint once, with the real alternatives: a private network or VPN, a separate public-access route guarded only by Jellyfin's token auth, or a TCP tunnel for devices that can run Pomerium CLI. It also covers the two things Jellyfin needs from the proxy to work correctly: the route has to preserve the host header so Jellyfin's absolute URLs and casting resolve, and Jellyfin has to list Pomerium under its Known Proxies. On identity, Jellyfin core has no SSO and will not trust a reverse-proxy header; a community plugin can add OIDC for the web client only, so for most setups the front-door gate is the practical ceiling. Because direct-play streams and downloads are long plain-HTTP responses, the example raises the total request timeout separately from WebSocket support so playback is not interrupted.

What ships is a runnable docker-compose (Pomerium and Jellyfin), a config with both Pomerium Zero console steps and Pomerium Core `config.yaml`, a simple request-flow diagram, and an access-channel table showing which clients Pomerium protects and which authenticate straight to Jellyfin. The bundled `validate/` directory is the sealed end-to-end harness for the example; with #2251 on main it runs in CI automatically, and it passed a full local run (Keycloak SSO login, live upstream assertions, network-isolation check) via `scripts/validate-guide-fixtures.sh jellyfin`.

This is one of four guides split out of the closed #2253 so each app can be reviewed on its own.

Closes [MAR-288](https://linear.app/pomerium/issue/MAR-288/use-pomerium-with-jellyfin-guide)

## AI assistance

Claude Opus drafted the guide prose, the docker-compose and config example, the diagram, and the validation fixtures. I reviewed and edited the content, doc-verified the client, Known-Proxies, and timeout claims against Jellyfin's documentation, and confirmed the site builds locally (`yarn build`, `cspell`, `prettier --check`). A follow-up cleanup commit (Claude-drafted, applying the review feedback on #2255 across the set) condensed duplicated framing, replaced the gating-matrix SVG with a markdown table, simplified the mermaid diagram, removed em and double-hyphen dashes per the guide style rules, scoped the prerequisites per tab, and fixed reviewer-caught accuracy gaps. I reviewed the cleanup before it was pushed; the local gates and the sealed E2E fixture (`yarn build`, `cspell`, `format-check`, `guide-audit`, `scripts/validate-guide-fixtures.sh`) were run by Claude on a merge with current main and passed.


content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+import Config from '/content/examples/guides/jellyfin/config.yaml.md';
+import Compose from '/content/examples/guides/jellyfin/docker-compose.yaml.md';
+
+# Secure Jellyfin with Pomerium
+
+## What this guide does
+
+You'll put a self-hosted [Jellyfin](https://jellyfin.org/) media server behind Pomerium so that Pomerium becomes the single front door for the web client: every browser request is authenticated against your identity provider (IdP) and checked against your policy before it ever reaches Jellyfin. Jellyfin keeps running its own login and user accounts on top, so Pomerium acts as an additional gate rather than replacing Jellyfin's accounts.
+
+The value of Pomerium here is centralized single sign-on (SSO), group-based policy, and an audit trail in front of a service you'd otherwise expose to the internet on its own. You sign in once with your organization identity, your group membership decides who can even reach the server, and every access decision is logged at the proxy. That is the point, not adding a second password prompt -- Jellyfin already has its own login.
```
kenjenkins: To me it feels like this paragraph duplicates the preceding one.

I'd suggest condensing it or removing it entirely. Or at least deleting the last "That is the point" sentence, which feels very LLM-y to me.

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+Replace `jellyfin.yourdomain.com` with your domain and `you@example.com` with the email (or switch to a group or domain match) that should be allowed through. Restart Pomerium after saving.
+
+</TabItem>
+</Tabs>
+
+## Configure Jellyfin
+
+Jellyfin needs to know its public URL and to trust the Pomerium hop so it does not mangle absolute URLs. The key settings in the Compose file below:
+
+- `JELLYFIN_PublishedServerUrl: https://jellyfin.yourdomain.com` -- Jellyfin builds absolute URLs (the web client, casting, Digital Living Network Alliance discovery) from this value, so it must match the public route host, not the container name.
```
kenjenkins: We can use real em dashes — no need to fake them using a double hyphen. (And likewise throughout this file.)

Also I'd suggest writing "From URL" rather than "public route host" to match the Pomerium-specific terminology above.

```suggestion
- `JELLYFIN_PublishedServerUrl: https://jellyfin.yourdomain.com` — Jellyfin builds absolute URLs (the web client, casting, Digital Living Network Alliance discovery) from this value, so it must match the **From** URL chosen above.
```

content/docs/guides/img/jellyfin/jellyfin-client-gating.svg
kenjenkins, line 1: A few comments on this chart:
1. I'm confused by some of the terminology. What is a "host-root Pomerium single sign-on (SSO) gate"? What does "gating the whole host" mean?
2. The color coding is odd. Why are 'yes' and 'no' both red? Should 'yes' be green? Why are the left column 'no's gray?
3. Something's gone wrong with the placement of the "breaks if host-gated" text.

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+# Secure Jellyfin with Pomerium
+
+## What this guide does
+
+You'll put a self-hosted [Jellyfin](https://jellyfin.org/) media server behind Pomerium so that Pomerium becomes the single front door for the web client: every browser request is authenticated against your identity provider (IdP) and checked against your policy before it ever reaches Jellyfin. Jellyfin keeps running its own login and user accounts on top, so Pomerium acts as an additional gate rather than replacing Jellyfin's accounts.
+
+The value of Pomerium here is centralized single sign-on (SSO), group-based policy, and an audit trail in front of a service you'd otherwise expose to the internet on its own. You sign in once with your organization identity, your group membership decides who can even reach the server, and every access decision is logged at the proxy. That is the point, not adding a second password prompt -- Jellyfin already has its own login.
+
+```mermaid
```
kenjenkins, line 31: I think this diagram may be more complicated than helpful.

I'm also confused about whether there's a recommended path for native apps or not? What is the "separate path" mentioned in the diagram?

If we do want a diagram here would something simpler be easier to understand? Maybe something with a structure more like this?

```
 Jellyfin
    ^
    |
 Pomerium
   ^  ^
  /    \
Web     Native app
browser
```

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+  J -->> U: Jellyfin web client (its own login on top)
+  Note over T,J: Native apps send Jellyfin API tokens<br/>and cannot complete the browser SSO gate
+  T ->> J: Direct API call, if you expose a separate path
+```
+
+## When to use this guide
+
+Use it when you run self-hosted Jellyfin and want to make sure only people from your organization can even reach the web interface, without exposing Jellyfin directly to the internet. Pomerium handles the network-level access decision; Jellyfin continues to manage libraries, its own user sessions, and playback behind that gate.
+
+This guide gates browser access. Before you adopt it, read the [Security considerations](#security-considerations): Jellyfin's native TV, mobile, and Chromecast apps authenticate directly to Jellyfin's application programming interface (API) with Jellyfin tokens and do not pass through a browser SSO flow, so gating the whole host root breaks them. If you rely on those clients, gate Jellyfin selectively or keep it on a private network reachable another way.
```
kenjenkins: I feel like this keeps repeating the same point, but in a vague way such that I still don't understand what it's trying to say.

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+## Prerequisites
+
+This guide assumes you've completed the [Quickstart](/docs/get-started/quickstart), so you already have Pomerium running and signing users in through the hosted authenticate service.
```
kenjenkins: Is the quickstart really needed? What does it cover that isn't included in the configuration files here?

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+## Run the stack
+
+The Compose file runs Pomerium Core and Jellyfin together. For Zero, drop the `pomerium` service and use the `compose.yaml` from the Quickstart with your `POMERIUM_ZERO_TOKEN`, keeping the `jellyfin` service below; put `jellyfin` on the same Docker network as the Quickstart's `pomerium` service (the Quickstart names it `main`) so Pomerium can resolve `jellyfin` by name.
+
+<Compose />
```
kenjenkins, line 91: The order seems off here. I'd expect this configuration file to fall under the "Configure Jellyfin" section heading, and the "After first boot" paragraph to follow the "docker compose up -d" command.

content/docs/guides/jellyfin.mdx
```diff
@@ -0,0 +1,164 @@
...
+2. **An allowed user reaches Jellyfin.** Sign in with a user your policy allows. Pomerium redirects you back and Jellyfin's own web client loads behind the gate.
+
+![The Jellyfin web client reached through Pomerium after single sign-on](./img/jellyfin/jellyfin-setup.png)
+
+3. **The upstream is actually live behind the gate.** With your session, open `https://jellyfin.yourdomain.com/System/Info/Public`. Jellyfin answers `200` with a small JSON document (it includes a `Version`); that endpoint needs no Jellyfin login, so a JSON response confirms the gate opened and the proxy reached a running Jellyfin rather than an error page.
+4. **The session is yours.** Open `https://jellyfin.yourdomain.com/.pomerium` to confirm your identity and the group claims Pomerium evaluated.
+
+Pomerium gates the route; Jellyfin runs its own login on top. Jellyfin's accounts and library setup are Jellyfin's concern, not Pomerium's.
+
+## Client surface and the gate
```
kenjenkins: This whole guide seems oddly obsessed with the phrase "the gate." Is this really the terminology we want to use?

desimone:
LLM' chewed up my manual edits when I moved from omnibus PR to this. Sorry, let me re-push after I manually review.

nickytonline (approved):
Went through the guide and everything works as expected. Ken's points were valid and have been addressed, so 👍🏻

kenjenkins (approved):

PR #2214, opened 2026-05-21
Session recording docs

kralicky:
## Summary

Adds docs for session recording.

## Related

https://linear.app/pomerium/issue/ENG-3932/docs-session-recording

Docs preview URL: https://deploy-preview-2214--pomerium-docs.netlify.app/docs/capabilities/session-recording

## AI disclosure

none

## Checklist

- [ ] reference any related issues
- [ ] updated docs
- [ ] updated UPGRADING.md
- [ ] updated CHANGELOG.md
- [ ] disclosed AI usage (or wrote "none") per AI_POLICY.md

PR #2266, opened 2026-06-11
docs: session recording doc improvements

nickytonline:
## Summary

Some copy cleanup in terms of explaining some acronymns and the bigger rework is giving the session recording auditing and session recording storage their own pages. I also changed the images to light theme as they show up better for docs.

preview page at https://deploy-preview-2266--pomerium-docs.netlify.app/docs/capabilities/session-recording

Snapzy_2026-06-18_17-01-08_157

## Related

Relates to #2214

## AI disclosure

Devin with Kimi K2.6 was used to seperate out the pages and update a bit of copy, e.g. expand acronyms. I reviewed the changes. I also made some manual changes including replacing dark mode screenshots with light themed ones as they look better.

## Checklist

- [x] reference any related issues
- [x] updated docs
- [ ] updated UPGRADING.md
- [ ] updated CHANGELOG.md


content/docs/capabilities/session-recording.mdx
```diff
@@ -976,6 +474,33 @@ This cache is shared with directory syncing for the Console.
...
+To stop recording new sessions, set `session_recording.enabled = false` on the route (or remove the `session_recording` block entirely). Existing recordings in storage are not affected.
+
+### Removing the extension
+
+To remove the session recording extension from a deployment:
+
+1. Disable recording on all routes.
+2. Remove the extension volume mount from your deployment manifest (Kubernetes, Docker Compose, or manual installation).
+3. Remove the extension path from `envoy_dynamic_extensions`.
+4. Restart Pomerium Core.
```
alexandreLamarre: restarts shouldn't be required

nickytonline: True. It should auto-reload config. 

```suggestion
```

content/docs/capabilities/session-recording/auditing.mdx
alexandreLamarre, line 1: I am not sure this belongs in internals since it describes setup of an end-to-end auditing feature available for session recording. I'd suggest condensing this as a proper nested tab of session-recording capability, similar to how we split apart the non-http capability tab. There may be other headers in the session-recording tab that make sense to split out as nested tabs as well, I'll leave that to you to make a judgement call

There are bits-and-bobs here and there in the auditing docs that are more reference style docs, but splitting it between capabilities & internals might be less obvious to operators (the reference style docs would not fit into our reference docs, since those are for configuration)

alexandreLamarre: Actually the auditing log fields may make sense in reference now that I think about it. We do have access log fields there as well.

alexandreLamarre (approved):

alexandreLamarre (approved):

PR #2273, opened 2026-06-22
docs: improve LLM markdown delivery and fix redirects

desimone:
This generates canonical Markdown sidecars for every non-API docs page and serves them as static files. Netlify serves `GET /docs/foo.md` directly, and the `llms.txt` family documents the convention (append `.md` to any `/docs/` page URL, excluding API pages), so agents get Markdown with no per-request runtime. The source stays in `content/`; the `.md` files are generated, served artifacts.

_Updated after review: an earlier revision added a Netlify edge function for same-URL `Accept: text/markdown` content negotiation. Per @kenjenkins and @nickytonline it is dropped, since the static `.md` files plus the documented append-`.md` convention cover agent discovery without running a function on every `/docs/*` request._

Three content changes ride along: the `llms.txt` family now distinguishes the Core, Zero, and Enterprise management models; generated links are normalized to canonical `www.pomerium.com` `.md` URLs; and HTML comments are stripped from generated Markdown.

It also fixes routing. `/docs/api` rewrote to `/docs/api.html`, which the build never produces, so it 404'd; it now rewrites to the built `/docs/api/pomerium-api`. The programmatic-access and upstream-mTLS redirects pointed at pages that have since moved, and `/docs/zero` gains a redirect to the current quickstart.

One check guards this: `yarn test:generated-links` resolves every generated link and redirect destination against a build. It's a standalone script (not wired into `pre-commit` or `docusaurus build`), validated locally here with a full `yarn build` plus the link check.

Known residual: Docusaurus reports a few pre-existing broken anchors (enterprise configure links, one MCP reference) unrelated to this change.

## AI assistance

Claude drafted the plugin, tests, and this description, and later removed the edge function in favor of static delivery after review. Bobby audited the full diff and verified the build, the link test, and the `/docs/api` fix manually before pushing.


nickytonline (commented):
Just wanted to call this out.

The edge function will run on every `/docs/*` request, not just agent traffic requesting markdown. Maybe fine, but not sure what Netlify plan we're on for docs. The Netlify free tier is I believe 500k edge function invocations/month.

desimone:
Dropped the edge function per your feedback — going with static `.md` delivery instead.

@kenjenkins Good call -- dropped the function. We serve the static `.md` files directly (Netlify serves them as plain files) and `llms.txt` documents the append-`.md` convention, so agents discover and fetch Markdown with no per-request function. The `<link rel="alternate">` tag stays a cheap follow-up if we ever want in-page discovery, but the static files plus llms.txt cover it today.

@nickytonline Removed the per-request function entirely, so there are zero added edge invocations -- the `.md` files are static and served directly. Thanks for flagging the plan/limit angle.

plugins/llms-txt-plugin.js
```diff
@@ -1,11 +1,109 @@
...
 const {visit, SKIP} = require('unist-util-visit');
 
+const CANONICAL_DOC_LINK_OVERRIDES = new Map([
+  ['/docs/guides', '/docs/guides'],
+  [
+    '/docs/capabilities/single-sign-out',
+    '/docs/capabilities/authentication#single-sign-out-sso',
+  ],
+  ['/docs/internals/cryptography', '/docs/internals/security#cryptography'],
+]);
```
kenjenkins: Why is this list needed? Do some of these belong in _redirects instead?

nickytonline (approved):

PR #2295, opened 2026-07-02
docs: fix stale version pins and typos

desimone:
Anyone copying the External-DNS, multi-replica, or IngressClass examples on the Kubernetes install page gets an ingress-controller pinned to v0.22.x — ten minor versions behind the v0.32.0 install command at the top of the same page. All three kustomization examples now pin v0.32.0.

Two smaller fixes ride along: the homelab pipeline guide listed Terraform provider `pomerium/pomerium v0.32.101`, a version that does not exist — now v0.32.0; and a typo plus a leftover cSpell suppression for a misspelling that no longer exists in the file.

Verified with: `yarn format-check && yarn cspell && yarn build` (all pass; `yarn cspell` runs `cspell --gitignore .`), `curl -sI https://raw.githubusercontent.com/pomerium/ingress-controller/v0.32.0/deployment.yaml` (200), and the Terraform registry versions API (`/v1/providers/pomerium/pomerium/versions` lists no 0.32.101; latest is 0.32.0).

**AI assistance:** AI (Claude) ran the docs-vs-code audit and drafted this diff; the version claims were verified with the commands above. Human review happens in this PR.

### DIFF
--- a/content/docs/deploy/k8s/install.md
+++ b/content/docs/deploy/k8s/install.md
@@ -78,7 +78,7 @@ kustomize build config/default
 
 ```yaml title="kustomization.yaml"
 resources:
-  - https://raw.githubusercontent.com/pomerium/ingress-controller/v0.22.1/deployment.yaml
+  - https://raw.githubusercontent.com/pomerium/ingress-controller/v0.32.0/deployment.yaml
 patches:
   - path: patch-proxy-external-dns.yaml
 ```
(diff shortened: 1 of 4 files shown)


content/docs/deploy/k8s/install.md
kenjenkins, line 1: Should we bump these to the latest patch release (v0.32.9) while we're at it?

kenjenkins (approved):

content/docs/deploy/k8s/install.md
desimone: Agreed — follow-up in #2303. It bumps all five pins on the page (the primary kubectl apply -k command plus the three example manifests and the multi-replica kustomize ref) to v0.32.9 so the page stays internally consistent; verified the deployment.yaml and config/default paths exist at that tag.

desimone: Update: went one better than the patch bump after a design question from Bobby — #2303 now points all install references at the 0-32-0 release branches (tip currently == v0.32.9), so the page tracks patch releases automatically and only needs touching at minor cuts. Also aligned the Gateway API CRD install to v1.4.1 per the controller's go.mod. Carries backport 0-32-0.

PR #2297, opened 2026-07-02
docs: fix DNS lookup family values and Zero policy examples

desimone:
Two pages state things the product contradicts.

The DNS reference documented a `DEFAULT` lookup-family value marked "Enterprise only". No such value exists: core accepts exactly `AUTO`, `V4_ONLY`, `V6_ONLY`, `V4_PREFERRED`, and `ALL` (`ValidateDNSLookupFamily` rejects anything else), none of them Enterprise-gated, and an unset `dns_lookup_family` defaults to `V4_PREFERRED`. The table now matches the code.

The Zero policies course page still said the `user` criterion "isn't an option in Zero" (a TODO from before it shipped). Zero's policy builder has it as **User ID** in the Criteria dropdown, so the TODO is now a real Example 3 that builds a user-ID policy. While fixing it we found the page's example YAML was invalid PPL — it put `is:` at the same level as the criterion key, which core's parser rejects with "each criteria may only contain a single key and value" — so Example 1's block is corrected to the nested form the PPL reference uses, and Example 3 follows it.

Verified with: `yarn format-check && yarn cspell && yarn build` (all pass); the DNS values and default against `config/options_dns.go`; every UI label in the new example against the Zero console source; both YAML forms against core's PPL parser (flat rejected, nested accepted).

**AI assistance:** implemented by Codex from a spec with code-evidence pointers, supervised and verified by Claude (every factual claim re-checked against core/Zero source, parser behavior proven with a live run). Human review happens in this PR.


content/docs/reference/dns.mdx
```diff
@@ -107,9 +107,10 @@ dns:
 
 ### Options
 
+If `dns_lookup_family` is unset or omitted, Pomerium applies `V4_PREFERRED`.
+
 | **Options** | **Description** |
 | :-- | :-- |
-| `DEFAULT` (Enterprise only) | Defers to configuration settings or default if none specified |
```
kenjenkins, line 112: Sadly there _is_ a 'Default' value in the Enterprise Console:

Screenshot 2026-07-02 at 10 15 45 AM

desimone: Good catch, thank you — that label is real. The Console's Select renderer maps the empty-string option to a "Default" label for DNS fields (ui/src/views/pages/Settings/components/Setting.tsx, getOptionLabel), which leaves dns_lookup_family unset; core then applies V4_PREFERRED. There is still no DEFAULT config value (ValidateDNSLookupFamily rejects it), so the latest commit documents both: the table keeps the five config values, and the prose above it now explains that the Console's Default entry means unset.

content/docs/reference/dns.mdx
```diff
@@ -107,9 +107,10 @@ dns:
 
 ### Options
 
+If `dns_lookup_family` is unset, Pomerium applies `V4_PREFERRED`. There is no `DEFAULT` config value: the **Default** entry in the Enterprise Console's DNS Lookup Family dropdown leaves the setting unset, with the same effect.
```
kenjenkins: I don't think the phrasing here makes sense. What if we move the sentence about 'Default' inside the 'Enterprise' tab above?

Then it could read something like this: "The 'Default' option leaves the setting unset, in which case the Core option setting will apply."

kenjenkins (approved):

content/docs/reference/dns.mdx
desimone: Done — moved into the Enterprise tab with your wording, and the Options intro now only states the unset default.

PR #2344, opened 2026-08-11
docs: fix reference table rendering in dark mode

wasaga:
## Summary

The Reference page's config table was unreadable in dark mode: light-gray text on a white grid surface.

The MUI theme in `src/theme/Root.js` was created without `palette.mode`, so every MUI component was permanently light-mode and the DataGrid kept a white background regardless of the site theme. `ReferenceTable.js` papered over this by forcing `rgba(224,224,224,1)` text in dark mode — onto that still-white surface.

`Root` renders outside Docusaurus' `ColorModeProvider` (Docusaurus renders `<Root><ThemeProvider>…`), so `useColorMode()` can't be called there — that's why the workaround existed.

Changes:

- `src/theme/muiTheme.js` (new) — `createAppTheme(mode)` builds the theme for a given color mode. The brand palette and `MuiDataGrid` style overrides moved here from `Root.js`. Dark mode uses the Infima dark primary (`#b296ff`, matching `src/css/custom.css`) and Docusaurus surface colors (`#1b1b1d` / `#242526`) so the grid blends with the page instead of using MUI's default `#121212`.
- `src/components/ReferenceTable.js` — nests its own `ThemeProvider` built from `useColorMode()`, which works there; drops the hardcoded text / sort-icon / menu-icon color overrides. The column-separator rule stays.
- `src/theme/Root.js` — uses the shared factory with an explicit light theme, plus a comment explaining why the mode isn't readable at that level.

Verified with `npm run build` + `docusaurus serve`, screenshotting `/docs/reference` in both modes: dark now has a dark grid surface, readable text, purple links and visible toolbar/sort/menu icons; light is unchanged except body text is full-contrast instead of the old 54%-opacity gray. `npm run format:check` passes.

Follow-up, not in this PR: other MUI consumers (homepage video cards, footer icon button) are still light-only. Fixing those site-wide would mean moving the `ThemeProvider` into a swizzled `Layout` wrapper, which renders inside `ColorModeProvider`.

## Related

n/a

## AI disclosure

Claude Code (Opus 5) diagnosed the root cause, wrote the change, and verified it by building the site and screenshotting both color modes.

## Checklist

- [ ] reference any related issues

### DIFF
--- a/src/components/ReferenceTable.js
+++ b/src/components/ReferenceTable.js
@@ -1,8 +1,10 @@
 import {useColorMode} from '@docusaurus/theme-common';
+import {ThemeProvider} from '@mui/material/styles';
 import {DataGrid, GridToolbar} from '@mui/x-data-grid';
-import React, {useState} from 'react';
+import React, {useMemo, useState} from 'react';
 
 import data from '../../content/docs/reference/reference.json';
+import createAppTheme from '../theme/muiTheme';
 import {renderCellExpand} from './RenderCellExpand';
 
 function filterHidden(item) {
@@ -16,6 +18,7 @@ export default function ReferenceTable() {
   });
 
   const {colorMode} = useColorMode();
+  const theme = useMemo(() => createAppTheme(colorMode), [colorMode]);
 
   const references = Object.values(data);
   const columns = [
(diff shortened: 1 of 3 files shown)


wasaga:
@calebdoxsey does it sound like a correct fix ?

calebdoxsey (approved):
I'm not sure what the correct thing to do for this is.

PR #2360, opened 2026-09-08
guides: remove duplicate config files

kenjenkins:
## Summary

Many of the Guides pages have example config files in the content/docs/examples/guides folder. Most of these have both .yaml and .yaml.md versions of the config. The plain .yaml files are used in automated tests, while the .yaml.md files are the ones actually included in the site source.

This opens up the possibility that the tested configurations could drift out of sync with the rendered guides.

Instead, let's use the webpack "raw loader" to include the plain .yaml files in the guide pages directly, using explicit <CodeBlock> components instead of the Markdown backtick fences (```).

## Related

n/a

## AI disclosure

Refactor performed by Claude Opus 4.5.

## Checklist

- [ ] reference any related issues


nickytonline:
@desimone can probably chime in here but I think there was two on purpose? One for actually testing the guide and one for content? That said, removing duplicates makes sense to me if the tests can still build with the configs.

nickytonline:
> @nickytonline if we have separate configuration files for the docs content and the tests, this opens up the possibility for the two to get out of sync. If this were to happen, then the tests would no longer verify that the docs content is correct. In contrast, if there is only one copy of the configuration files, then the tests should accurately verify the correctness of the docs content.
> 
> 
> 
> But please do let me know if I'm misunderstanding something about the setup here.

I'm definitely pro one config for each, I just thought there was a reason we needed to have duplicates for the tests, that's all.

We're probably good to merge this if all checks pass.

nickytonline (approved):

